PR #25, opened 2017-05-17
DOC-459: Tooltips for Dashboard mysql-query-response-time

azhebel:


dashboards/MySQL_Query_Response_Time.json
```diff
@@ -44,6 +45,7 @@
             "type": "dashboards"
         }
     ],
+    "refresh": "1m",
```
michaelcoburn: @azhebel This was set to false before, do we need to maintain that? @askomorokhov , @rnovikovP ?

dashboards/MySQL_Query_Response_Time.json
```diff
@@ -719,13 +760,15 @@
...
                 "tagsQuery": null,
-                "type": "query"
+                "type": "query",
+                "useTags": false
             }
         ]
     },
     "time": {
-        "from": "now-12h",
+        "from": "now-1h",
```
michaelcoburn: @azhebel I think we should ensure this remains now-12h

dashboards/MySQL_Query_Response_Time.json
askomorokhov: Yes, we do. Here are the rules that we must follow: https://confluence.percona.com/display/SER/Grafana+dashboards+rules

michaelcoburn: @askomorokhov ok, thanks for sharing, I wasn't aware we had a style guide! :)  Does this mean that @azhebel should re-submit his PR? Or what is our next step now to move this PR forward?

askomorokhov: Fixed. 🙂

askomorokhov:
Looks good!

PR #32, opened 2017-06-09
Use UTC by default in all dashboards.

jsha:
This makes things much simpler for teams coordinating across timezones, and is
also more likely to match server log files (which are typically recorded in
UTC).

### DIFF
--- a/dashboards/Amazon_RDS_OS_Metrics.json
+++ b/dashboards/Amazon_RDS_OS_Metrics.json
@@ -1131,7 +1131,7 @@
             "30d"
         ]
     },
-    "timezone": "browser",
+    "timezone": "utc",
     "title": "Amazon RDS OS Metrics",
     "version": 0
 }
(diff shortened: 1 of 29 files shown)


rnovikovP:
Can you use Settings changes for timezone in Grafana? Because in UI usually 'browser time'. 
Also, this can be an option completely remove this setting from json and use the Default setting of time. ?

jsha:
I can change in Settings, but when using the recommended install method (Grafana loads from JSON files on disk), it is impossible to save changes to the dashboard, so my team has to change the Settings every time they load the dashboard.

> Also, this can be an option completely remove this setting from json and use the Default setting of time. ?

You mean, have a time display setting on a user basis rather than a dashboard basis? I think that is not currently possible, but it would be a good upstream feature request for Grafana. I will file it.

jsha:
Ah, looks like this issue in Grafana is relevant: https://github.com/grafana/grafana/issues/8503. If you set a dashboard's timezone to "default", that is evaluated once and then saved to a specific timezone. However, for JSON dashboards (as opposed to DB dashboards), that might not be the case? I will try with a "default" setting.

AlekSi:
@jsha did "default" setting work?

jsha:
Yes, it worked. Thanks!

PR #37, opened 2017-06-19
Add Travis CI integration

AlekSi:
PMM-963


AlekSi:
@askomorokhov PTAL. Do we need those Python scripts to be compatible with something?

delgod:
@AlekSi these scripts are used during manual exporting/importing from/to pmmdemo.percona.com

PR #54, opened 2017-09-25
PMM-1508: check that master UUID is present.

AlekSi:
Single stat panel works only with single metric.


AlekSi:
According to @michaelcoburn, 5.5 doesn't have `master_uuid`, that fix will break dashboard for it.

PR #79, opened 2017-11-25
PMM-1761 Display only hosts with Mysql and MongoDB

rnovikovP:
I've used expression we have for Mysql and MongoDB overview boards.  Later will be replaced by mongo_up


askomorokhov:
> Wait. Shouldn't dist and src be equal?
@AlekSi, yes - but we need to do changes in src - then build process must create dist dir.

peterschwaller:
touch

PR #99, opened 2018-01-24
PMM-1965 Fix drill down links -> Prometheus Exporters Overview

adivinho:


dashboards/Prometheus_Exporters_Overview.json
```diff
@@ -1335,5 +1341,5 @@
     },
     "timezone": "browser",
     "title": "Prometheus Exporters Overview",
-    "version": 12
+    "version": 13
```
rnovikovP: just to clarify: we don't use versions right now 
but, we can use it just for usability reason and let us easily understand what version  installed

PR #108, opened 2018-02-01
PMM-1511 PXC cluster not recognized correctly

adivinho:
PXC cluster not recognized correctly if a MariaDB node is present in
the list of hosts


peterschwaller:
touch

PR #121, opened 2018-02-13
PMM-2034 fix broken symlink in dest dir

delgod:


peterschwaller:
Dummy comment to trigger sync to Jira.  Please ignore.

PR #140, opened 2018-02-23
PMM-2010 adopt network and disk latency graphs

delgod:


rnovikovP:
@adivinho  can you resolve this conflict and apply changes to our new Home  (renamed) dashboard also. Thanks

delgod:
@rnovikovP I rebased my changes, no conflicts right now

adivinho:
- Combined metrics with an interval
- Added rds nodes to the hosts list for the "Disk Performance" dashboard

ArvinDevel:
hi, I'm a new bee to Prometheus. I wonder why use complicated `rate(node_network_receive_bytes{xxx}[5m]) and irate(node_network_receive_bytes{xxx}[5m]) or xxx` query when calculate the network rate. To my knowledge, the value before `or` will be the final value when it has value, because `rate` and `irate` all has the same metrics label. Is this a way to avoid failure when `node_network_receive_bytes` not exists?

adivinho:
Hello Arvin,
You are right it's a workaround to remove gaps in graphs when a display interval is lower than a scraping interval.

PR #141, opened 2018-02-23
PMM-2109-renamed-Home-to-Home-Dashboard

adivinho:


delgod:
please also do the same in the import script https://github.com/percona/pmm-server/blob/master/import-dashboards.py#L223-L243

PR #154, opened 2018-03-16
PMM-781 Plot new PXC status variables on new graph

adivinho:


rnovikovP:
LGTM. but  see some comments in Jira

PR #180, opened 2018-04-25
import mysql.json dashboard error

haozi4263:
hi
i use grafana import mysql-dashboard error

error log:
Datasource named Prometheus was not found


rnovikovP:
Sorry, looks like there some error in this Pool request.  

If there some Bug in PMM -please submit in in our Jira.

The  separate scripts are not  tested outside PMM anyway

PR #239, opened 2018-08-28
PMM-3166 ProxySQL_Overview changes

impimp:
- removed `Status number to state mapping' panel,
 - added value to text conversion in `Endpoint Status'.


rnovikovP:
@impimp  thanks!  this will be in 1.16.  release

PR #252, opened 2018-10-03
PMM-3027 Fix queries with RDS vCPUs.

AlekSi:


AlekSi:
Wait :) There is a bug

dashboards/Home_Dashboard.json
```diff
@@ -2210,7 +2210,7 @@
             "tableColumn": "",
             "targets": [
                 {
-                    "expr": "count(node_cpu{mode=\"user\", instance=~\"$host\"}) or count(rdsosmetrics_General_numVCPUs{instance=~\"$host\"})",
+                    "expr": "count(node_cpu{instance=~\"$host\",mode=\"user\"}) or rdsosmetrics_General_numVCPUs{instance=~\"$host\"}",
                     "format": "time_series",
                     "instant": true,
                     "interval": "5m",
```
AlekSi: @adivinho Why `5m` and not `$interval`?

dashboards/Home_Dashboard.json
```diff
@@ -1160,7 +1160,7 @@
             "tableColumn": "",
             "targets": [
                 {
-                    "expr": "count(node_cpu{instance=~\"$host\",mode=\"user\"} or rdsosmetrics_General_numVCPUs{instance=~\"$host\"})",
+                    "expr": "count(node_cpu{instance=~\"$host\",mode=\"user\"}) or rdsosmetrics_General_numVCPUs{instance=~\"$host\"}",
```
adivinho: IMHO changes have to be reverted. In this variant we will ignore RDS CPUs

dashboards/Home_Dashboard.json
adivinho: AFAIR It was done in order to slightly reduce time of the dashboard loading.

dashboards/Home_Dashboard.json
AlekSi: Why?

AlekSi: Both variants are wrong. 😆 It should be
```
count(node_cpu{instance=~"$host",mode="user"}) + sum(rdsosmetrics_General_numVCPUs{instance=~"$host"})
```

I will send a patch.

AlekSi: … plus it needs to be adjusted for the case when no RDS instances are monitored and `sum` returns no data

AlekSi: … and this is wrong too because `node_cpu` does not exist for RDS hosts

adivinho: It works ...
screenshot_20181004_164501

AlekSi:
This is still wrong…

AlekSi:
Do not merge. This does not work: https://jira.percona.com/browse/PMM-3027?focusedCommentId=230006

dashboards/Home_Dashboard.json
adivinho: Please change mode=\"user\" to mode=~\"user\". Otherwise the first part of the formula will return mode as an element in reply. It will cause a conflict with a reply from the second part because there will be no elements.

AlekSi: I can confirm that: https://pmmdemo.percona.com/prometheus/graph?g0.range_input=1h&g0.expr=1-absent(node_cpu%7Binstance%3D%22rds-mysql56%22%2Cmode%3D%22user%22%7D)&g0.tab=1&g1.range_input=1h&g1.expr=1-absent(node_cpu%7Binstance%3D%22rds-mysql56%22%2Cmode%3D~%22user%22%7D)&g1.tab=1

But why? What part of the PromQL triggers that behavior?

AlekSi: https://prometheus.io/docs/prometheus/latest/querying/functions/#absent()

> In the second example, absent() tries to be smart about deriving labels of the 1-element output vector from the input vector.

🤢

adivinho: Please set mode=~\"user\ in the first formula too.
It's possible that only an RDS host is selected in hosts drop-down menu.

AlekSi: Good catch. Wrong copy&paste.

AlekSi:
PTAL

PR #266, opened 2018-10-26
PMM-2928 Query Response Time Add to Dashboard

adivinho:


rnovikovP:
let's use https://github.com/percona/grafana-dashboards/pull/302  only for 2.0

PR #269, opened 2018-11-01
PMM-3175 create dashboards to reflect memory metrics collected by node exporte

adivinho:


rnovikovP:
we'll use  https://github.com/percona/grafana-dashboards/pull/301 instead

PR #322, opened 2019-03-08
PMM-3651 modify cleanup-dash.py script refer to changed in Grafana 6.0

adivinho:


misc/cleanup-dash.py
```diff
@@ -170,6 +170,9 @@ def drop_some_internal_elements(dashboard):
             del dashboard['__inputs']
         if '__requires' in element:
             del dashboard['__requires']
+        for list_index, listelement in enumerate(dashboard['templating']['list']):
```
askomorokhov: `list_index` - is unused

misc/cleanup-dash.py
```diff
@@ -170,6 +170,9 @@ def drop_some_internal_elements(dashboard):
...
+            if 'current' in listelement:
+                listelement['current'] = {}
```
askomorokhov: Isn't it more clear `dashboard['templating']['list'][index] = {}` ?

adivinho: Dropping of whole section list is leading to next error.
So I have rewritten it to
dashboard['templating']['list'][index]['current'] = {}

Screenshot_20190311_152430

PR #323, opened 2019-03-11
PMM-3165 Remove Consul from PMM Server

adivinho:


.travis.yml
```diff
@@ -8,8 +8,8 @@ before_script:
   - docker --version
   - docker-compose --version
   - docker-compose up -d
-
   - sudo pip install --upgrade requests
+  - sudo pip install --upgrade pip
```
AlekSi: We definitely should upgrade pip before using it to install other packages. 😆 Swap those lines.

adivinho: Done.

PR #503, opened 2019-11-10
PMM-4890: grafana dashboards variables passing

lunaticusgreen:


pmm-app/src/pmm-qan-app-panel/module.ts
```diff
@@ -161,36 +168,50 @@ export class PanelCtrl extends MetricsPanelCtrl {
...
+      'var-agent_id',
+      'var-service_id',
+      'var-az',
+      'var-node_type',
+      'var-node_model',
+      'var-node_name',
+      'var-service_name',
+      'var-service_type',
+      'var-username',
+    ];
```
askomorokhov: Could you please sort/group them in a similar way (for easier review and do not accidentally lose any of them):
https://github.com/percona/qan-api2/blob/master/services/analytics/base.go#L121

PR #537, opened 2019-12-10
PMM-5061: inventory refactoring

lunaticusgreen:


pmm-app/src/pmm-inventory/DataService.ts
```diff
@@ -0,0 +1,88 @@
...
+  static generateStructure(item) {
+    const addAgentType = Object.keys(item).map(agentType => new Object({ agentType: agentType, params: item[agentType] }));
+    const createParams = addAgentType.map(agent =>
+      agent['params'].map(arrItem => new CommonModel(arrItem, InventoryDataService.getType(agent['agentType'])))
+    );
+    return [].concat(...createParams);
+  }
+
+  static getType(type) {
+    switch (type) {
```
askomorokhov: @lunaticusgreen, what do you think to change this `switch/case` to the next form:

```suggestion
    typeMap = {
       external_exporter: 'External exporter',
       mongodb_exporter: 'MongoDB exporter'
       ....
    }
    return typeMap[type] || ''
```

I know, it is a more pythonic way, but at least for me looks more compact and readable.

pmm-app/src/pmm-inventory/panel.tsx
```diff
@@ -0,0 +1,208 @@
...
+import './panel.scss';
+import InventoryService from './InventoryService';
+import { InventoryDataService } from './DataService';
+
+const { TabPane } = Tabs;
+
+const getCustomLabels = customLabels => {
+  return customLabels.map(label => (
+    <span>
+      `${label.key}: ${label.value}` }
```
askomorokhov: Is it intended to leave trailing `}`?
```suggestion
      `${label.key}: ${label.value}`
```

askomorokhov: or 
```suggestion
     { `${label.key}: ${label.value}` }
```

pmm-app/src/pmm-inventory/DataService.ts
lunaticusgreen: Well, pythonic sounds good enough for me :D 
Switch here is legacy after migration this code from qan-app and simplification, I thought that we may need it in case if we need to add some additional logic here, but I think you're right, we just can add it if we need it

pmm-app/src/pmm-inventory/panel.tsx
lunaticusgreen: Yes, typo, fixed, thanks =)

lunaticusgreen:
How it looks now: 
image

PR #544, opened 2019-12-17
PMM-5133: node type inventory

lunaticusgreen:
https://github.com/Percona-Lab/pmm-submodules/pull/576


lunaticusgreen:
image

pmm-app/src/pmm-inventory/panel.tsx
```diff
@@ -94,6 +94,11 @@ export const nodesColumns = [
     dataIndex: 'node_id',
     key: 'node_id',
   },
+  {
+    title: 'Node Type',
```
BupycHuk: looks like something wrong here. Title is `Node Type`, but key and dataIndex are `agentType`

lunaticusgreen: This is just a common name for this type of column, all tables have similar structure

BupycHuk: maybe it's better to rename it then. looks really confusing.

PR #557, opened 2019-12-24
PMM-5206: dashboards styling fixed

lunaticusgreen:
https://github.com/Percona-Lab/pmm-submodules/pull/592

For some mysterious reason we had a problem with panel-container styling after navigation from left grafana menu, to prevent it in the future, panel-container wrapper will be styled now with special class - custom-grafana-plugin

### DIFF
--- a/pmm-app/src/pmm-settings/panel.tsx
+++ b/pmm-app/src/pmm-settings/panel.tsx
@@ -43,7 +43,7 @@ const SettingsPanel = () => {
       <div className={'app-theme-dark pmm-settings-panel'} style={{ width: '100%' }}>
         <FormFinal
           onSubmit={() => {}}
-          render={(lin: ReactElement => {
+          render={(): ReactElement => {
             // @ts-ignore
             return (
               <Collapse bordered={false} defaultActiveKey={['1']} onChange={() => {}} style={customCollapseStyle}>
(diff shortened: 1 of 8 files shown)


pmm-app/src/pmm-settings/panel.tsx
```diff
@@ -28,12 +29,13 @@ const customCollapseStyle = {
 };
 
 const SettingsPanel = () => {
+  // document.getElementById('antd').classList.add('custom-grafana-plugin');
```
BupycHuk: do we need it?

pmm-app/src/react-plugins-deps/components/helpers/styling.ts
```diff
@@ -0,0 +1,16 @@
+class Styling {
+  static setPluginPanelClass(panelWrapperId = '#antd', newClass = 'custom-grafana-plugin') {
```
BupycHuk: maybe `panelWrapperSelector`?

BupycHuk: looks like `addPluginPanelClass` fits better for this method. Set usually means replace old value with a new one.

puneet0191:
Ready for Merge

PR #573, opened 2020-01-21
PMM-5340 fix graph scraping time drift

adivinho:
- [ ] https://github.com/Percona-Lab/pmm-submodules/pull/646

### DIFF
--- a/dashboards/Prometheus.json
+++ b/dashboards/Prometheus.json
@@ -36,7 +36,7 @@
     "gnetId": null,
     "graphTooltip": 1,
     "id": null,
-    "iteration": 1577450608139,
+    "iteration": 1579609876117,
     "links": [
         {
             "icon": "doc",
@@ -3599,14 +3599,14 @@
                         "Rushed Mode High Watermark": "#BF1B00",
                         "Rushed Mode Low Watermark": "#EAB839"
                     },
-                    "bars": true,
+                    "bars": false,
                     "dashLength": 10,
                     "dashes": false,
                     "datasource": "Prometheus",
                     "decimals": 2,
                     "editable": true,
                     "error": false,
-                    "fill": 0,
+                    "fill": 6,
                     "fillGradient": 0,
                     "grid": {},
                     "gridPos": {
@@ -3630,10 +3630,10 @@
                         "total": false,
                         "values": true
                     },
-                    "lines": false,
+                    "lines": true,
                     "linewidth": 1,
                     "links": [],
-                    "nullPointMode": "null",
+                    "nullPointMode": "null as zero",
                     "options": {
                         "dataLinks": []
                     },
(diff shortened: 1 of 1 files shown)


dashboards/Prometheus.json
```diff
@@ -3647,14 +3647,11 @@
                     "steppedLine": false,
                     "targets": [
                         {
-                            "expr": "time() - node_time_seconds",
-                            "format": "time_series",
+                            "expr": "clamp_min((time() - max by (node_name) (node_time_seconds)) - scalar(floor(prometheus_target_interval_length_seconds{quantile=\"0.99\",interval=~\"$scrapingInterval\"})),0)",
```
AlekSi: Why `floor()` and not `round()`? And why round to integer seconds at all?

dashboards/Prometheus.json
```diff
@@ -5575,6 +5573,28 @@
...
+                "definition": "label_values({__name__=~\"node_time_seconds\"},job)",
+                "hide": 2,
+                "includeAll": false,
+                "label": null,
+                "multi": false,
+                "name": "scrapingInterval",
+                "options": [],
+                "query": "label_values({__name__=~\"node_time_seconds\"},job)",
+                "refresh": 2,
+                "regex": "/.*_hr-([0-9]m[0-9]s|.?[0-9]s).*/",
```
AlekSi: To be clear – we can't rely on that job name format in a long run. It _will_ be broken in the future.
/cc @rnovikovP

dashboards/Prometheus.json
adivinho: Thank you. Round() is better. 

Screenshot_20200122_132727

dashboards/Prometheus.json
```diff
@@ -3647,14 +3647,11 @@
                     "steppedLine": false,
                     "targets": [
                         {
-                            "expr": "time() - node_time_seconds",
-                            "format": "time_series",
+                            "expr": "clamp_min((time() - max by (node_name) (node_time_seconds)) - scalar(round(prometheus_target_interval_length_seconds{quantile=\"0.99\",interval=~\"$scrapingInterval\"})),0)",
```
AlekSi: The question remains: why round to the integer at all? why not to one or two digits after dot? If I understand correctly, that will make a graph a bit more smooth.

dashboards/Prometheus.json
adivinho: IMHO the only way to omit using such approach is splitting current job parameters into two parts:
job="node_exporter_agent_id_bcd0b2b6-f063-4d55-b51c-4445e6ee6165_hr"
interval="5s"

Also I don't know any other dashboard element that would require an interval parameter.

e.g.
prometheus_target_interval_length_seconds{instance="pmm-server",interval="1m0s",job="prometheus",quantile="0.99"}

node_time_seconds{...,job="node_exporter_agent_id_bcd0b2b6-f063-4d55-b51c-4445e6ee6165_hr-5s",...}

dashboards/Prometheus.json
adivinho: Let's simplify it by removing a rounding function completely.

dashboards/Prometheus.json
rnovikovP: re "It will be good to have additional metrics": let's add them to appropriate Issues in Jira to  not lost. But as I can see it's not blocking us from workaround it for now?

PR #574, opened 2020-01-22
PMM-5065: alert manager UI

lunaticusgreen:
Code returned and some integration issues fixed:
https://github.com/Percona-Lab/pmm-submodules/pull/645


pmm-app/src/react-plugins-deps/components/validators/validators.ts
```diff
@@ -6,6 +6,14 @@ class Validators {
     }
     return 'Port should be a number and between the range of 0 and 65535';
   }
+
+  static validateIP(value) {
```
AlekSi: We do server-side validation.
And that function seems unused.

pmm-app/src/pmm-settings/Parts/AlertManager/AlertManager.test.tsx
```diff
@@ -8,7 +8,7 @@ describe('Settings Part test', () => {
     const component = renderer.create(
       <AlertManager
         settings={{
-          alert_manager_ip: '192.168.0.1',
+          alert_manager_url: '192.168.0.1',
```
AlekSi: That's not an URL. It will not work. Please actually test it.

lunaticusgreen: Updated it now, I'll also change this tests, because now this fields is not really matters, we don't check it

AlekSi:
But please do not merge it yet – we do not want it in 2.2.2.

pmm-app/src/react-plugins-deps/components/helpers/Helpers.tsx
```diff
@@ -4,7 +4,7 @@ import React, { ReactComponentElement } from 'react';
 interface VerticalFormWrapperInterface {
   alignLabel?: string;
   label: string;
-  tooltip: string | JSX.Element;
+  tooltip?: string | JSX.Element;
```
atymchuk: Better to use `React.ReactNode`: it's a union of string & React elements

PR #576, opened 2020-01-22
PMM-5210: data retention units changed to days

lunaticusgreen:
https://github.com/Percona-Lab/pmm-submodules/pull/648


lunaticusgreen:
This code is already included in https://github.com/percona/grafana-dashboards/pull/594 @Nailya is already testing it, so we can close this PR

PR #587, opened 2020-02-07
PMM-5406: grafana tookit update

lunaticusgreen:
Grafana supports LESS extension now, so we can remove our little fork


atymchuk:
@lunaticusgreen We have to fix the builds

PR #605, opened 2020-02-25
PMM-5305: time range change fix

lunaticusgreen:


pmm-app/src/pmm-settings/Parts/Settings/Settings.tsx
```diff
@@ -176,20 +195,20 @@ const SettingsPart = props => {
...
-                          options={dataRetentionOptions}
-                          defaultValue={'weeks'}
-                          style={{ width: '40%' }}
-                        />
-                      </div>
+                      <InputField
+                        name={'data_retention_count'}
+                        validate={Validators.compose(
+                          Validators.range(1, 3650),
+                          Validators.required
```
atymchuk: `required` should come first, anything else should almost always follow, not vice versa.
It just doesn't make much sense to check the value if there in no value.
See an example [here](https://github.com/final-form/react-final-form/blob/master/examples/field-level-validation/index.js#L55).

pmm-app/src/pmm-settings/Parts/Settings/Settings.tsx
```diff
@@ -69,6 +69,25 @@ export const getMetricsResolutionValues = metricsResolutions => {
   return metricsIndex;
 };
 
+const SECONDS = 60;
```
atymchuk: It should be all in seconds, because that's what the API is looking for.
MINUTE  = 60;
HOUR = MINUTE * 60;
DAY = HOURS * 24;
Then the rest will make sense.

PR #624, opened 2020-03-09
PMM-5541: details error

lunaticusgreen:


pmm-app/docker-compose.yaml
```diff
@@ -0,0 +1,144 @@
+version: '3'
+services:
+  pmm-server:
+    container_name: pmm-server
+    image: perconalab/pmm-server:dev-latest
+    volumes:
+      - "/Users/lunaticus/Documents/Work/Percona/grafana-dashboards/pmm-app:/var/lib/grafana/plugins"
```
atymchuk: We should avoid hard-coded configs, I suggest we use an environment variable.
```yml
    volumes:
      - "${PMM_APP_PATH}:/var/lib/grafana/plugins"
```

lunaticusgreen: Thanks! this line got here by accident

pmm-app/docker-compose.yaml
```diff
@@ -0,0 +1,144 @@
...
+              --mysql-db=sbtest \
+              --mysql-table-engine=innodb \
+              --oltp-reconnect \
+              --oltp-table-size=1000000 \
+              /usr/share/sysbench/tests/include/oltp_legacy/select.lua \
+              run
+      "
+
+  sysbench-pstpcc:
+    image: perconalab/sysbench
```
atymchuk: I would add a container name everywhere for consistency sake, like: `container_name: sysbench-pstpcc`

lunaticusgreen: We actually don't really need it here, because first of all this configuration will be reworked soon, and I use docker-compose configuration taken from backend developers, any unnecessary will lead to more complex comparing with that file.

PR #658, opened 2020-04-19
SAAS-31 database checks UI MVP

atymchuk:
- [ ] https://github.com/Percona-Lab/pmm-submodules/pull/798


atymchuk:
> Looks good, but could be better if add e2e tests and maybe some unit tests for the sanity check

Sure, will do next thing.

pmm-app/package.json
```diff
@@ -11,7 +11,7 @@
   "main": "module.js",
   "scripts": {
     "build": "grafana-toolkit plugin:build",
-    "test": "grafana-toolkit plugin:test --updateSnapshot",
+    "test": "grafana-toolkit plugin:test",
```
lunaticusgreen, line 14: Good fix, but we'll also need command to update snapshots, so please add additional script for it

atymchuk: I don't get it: the way it was before, `npm test` would pass even if the snapshots are different. I'm sure that's not the way snapshots have been designed.

atymchuk: Anyway, added another script with --updateSnapshot

PR #666, opened 2020-04-25
PMM-5830: cleanup form dependencies

lunaticusgreen:


pmm-app/src/pmm-add-instance-app-panel/AddInstance/AddRemoteInstance/AddRemoteInstance.tsx
```diff
@@ -213,32 +203,31 @@ const AddRemoteInstance = props => {
               <span className="description">Port your service is listening on</span>
             </div>
             <div className="add-instance-panel">
-              <InputField form={form} name="username" placeholder="Username" required={true} />
+              <InputField name="username" placeholder="Username" required={true} />
```
atymchuk, line 206: `true` is the default value for props passed to components, so syntax should be just `required`. There is also a widely adopted [rule](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-boolean-value.md), which I encourage to use.

pmm-app/src/pmm-settings/Parts/AlertManager/AlertManager.tsx
```diff
@@ -77,12 +73,7 @@ const AlertManager = props => {
                   />
                 }
                 element={
-                  <InputField
-                    form={form}
-                    name="alert_manager_url"
-                    placeholder="Enter URL"
-                    style={{ width: '100%' }}
-                  />
+                  <InputField name="alert_manager_url" placeholder="Enter URL" style={{ width: '100%' }} />
```
atymchuk, line 76: Better to use emotion for styling. Also, we should prefer classes over inline styles.

pmm-app/src/pmm-add-instance-app-panel/AddInstance/AddRemoteInstance/AddRemoteInstance.tsx
lunaticusgreen: Thanks, great comment, I'll add it to eslint

PR #671, opened 2020-04-29
PMM-5767 Adding Selenoid_helper back 

puneet0191:


pmm-app/tests/pages/dashboardPage.js
```diff
@@ -370,19 +370,20 @@ module.exports = {
...
+    let sections;
     if (typeof sectionsToExpand === 'string') {
       sections = [sectionsToExpand];
     } else {
       sections = sectionsToExpand;
     }
     for (let i = 0; i < sections.length; i++) {
-      const sectionName = sections[i].toString().split('(');
-      const rowToExpand = `${this.fields.collapsedDashboardRow}[contains(text(), '${sectionName[0]}')]`;
+      let sectionName = sections[i].toString().split('(');
```
atymchuk, line 380: @puneet0191 Is there a reason to drop `const` in favor of `let`?

puneet0191: @atymchuk I think I keep getting this: `Missing initializer in const declaration` so I wanted to avoid it.

atymchuk: That's because your IDE is not properly set up.

puneet0191: I was getting it while running the tests

PR #695, opened 2020-05-06
Hotfix for PMM-5878 and PMM-5879

lunaticusgreen:
I used timeout because for some reason right after load jquery parents is not available, If you know a better way - feel free to apply it.


pmm-app/src/pmm-qan-app-panel/module.ts
```diff
@@ -79,73 +79,76 @@ export class PanelCtrl extends MetricsPanelCtrl {
...
+          'showSuccessNotification',
+          () => {
+            AppEvents.emit('alert-success', ['Content has been copied to clipboard']);
+          },
+          false
+        );
+
+        frame
+          .contents()
+          .bind('DOMSubtreeModified', () => setTimeout(() => $scope.ctrl.calculatePanelHeight(), 10));
```
askomorokhov, line 149: Do we need these nested timeouts now?

pmm-app/src/pmm-qan-app-panel/module.ts
```diff
@@ -79,73 +79,76 @@ export class PanelCtrl extends MetricsPanelCtrl {
...
+        $scope.qanParams.type,
+        $scope.qanParams.search,
+        $scope.qanParams.filters,
+        $scope.qanParams.main_metric,
+        $scope.qanParams.columns,
+        $scope.qanParams.order_by,
+        $scope.qanParams.group_by,
+        $scope.qanParams.filter_by,
+        $scope.qanParams.active_details_tab,
+      ] = this.retrieveDashboardURLParams();
```
askomorokhov, line 128: Can we move this assignment out to parent function?

atymchuk: Maybe not worth the effort, since Angular goes away anyway.

PR #706, opened 2020-05-09
PMM-5903: refresh fix

lunaticusgreen:


lunaticusgreen:
@askomorokhov You are the only one who worked with it, please look at these changes, It's workaround for page reloading that appeared after grafana update.

PR #716, opened 2020-05-13
PMM-5935: change tooltips

lunaticusgreen:


pmm-app/src/pmm-settings/Parts/Settings/Settings.tsx
```diff
@@ -160,8 +160,8 @@ const SettingsPart = props => {
                     links={[
                       {
                         url:
-                          'https://www.percona.com/doc/percona-monitoring-and-management/' +
-                          '2.x/faq.html#what-resolution-is-used-for-metrics',
+                          'https://www.percona.com/doc/percona-monitoring-and-management' +
```
atymchuk, line 163: It looks like we are using one and the same string constant in different places: `https://www.percona.com/doc/percona-monitoring-and-management/2.x/manage/server-admin-gui.html`.
Why don't we save it (const docUrl = '') to a file in `pmm-app/src/pmm-settings/Parts` and import it from there?

lunaticusgreen: Good point, thanks, moved it to constants file

PR #723, opened 2020-05-17
PMM-5937 db check panel remake

atymchuk:
- [ ] https://github.com/Percona-Lab/pmm-submodules/pull/852

* no more antd, its Table is replaced by a custom one
* made the panel transparent to get rid of panel style overrides
* reduced bundle size from 2.35Mb to just 55Kb
* unit test coverage


atymchuk:
@puneet0191 I guess the right way to fix codecept is to lock its version in `devDependencies`:
`"codeceptjs": "2.6.2"`

puneet0191:
@atymchuk sure you are right but we are using Codeceptjs docker image here, I used docker image version, so may be we need to put it in config somewhere

atymchuk:
Closing this PR, since PMM-5936 fully embodies this one.

PR #727, opened 2020-05-19
PMM-5796 no imperative styles

atymchuk:
FB - https://github.com/Percona-Lab/pmm-submodules/pull/858
* fix all imperative styles on three panels
* made the unit tests look clean again (!)
* fixed the jumping centered table bug in PMM inventory
* fixed e2e tests


pmm-app/src/react-plugins-deps/components/FormComponents/FormElement/FormElement.tsx
```diff
@@ -45,7 +45,7 @@ const HorizontalFieldLayout = ({ label, tooltip, element }) => {
     <>
       <Row align="middle">
         <Col span={8} className={style.fieldLabelColumn}>
-          <strong>{label}</strong>
+          <span>{label}</span>
```
atymchuk, line 48: This was bold, but the headers were normal.

atymchuk:
> I'm not sure that changing field labels and tooltip text is a great idea when you do refactoring, but anyway, great cleanup and readability improvement!

Generally speaking, maybe not. On the other hand,  it is a great idea if you apply the rule "Leave the code better than you found it".
Think how much effort it would take to do three separate pull requests instead of just one.

PR #728, opened 2020-05-20
PMM-5524 Fixing tests

puneet0191:


AlekSi:
But shouldn't we create a FB? ;)

puneet0191:
I just ran the UI test job with this branch and it worked fine, so I guess FB is not needed :smile:

PR #736, opened 2020-05-22
SAAS-152  enabled custom locator plugin, refactored selectors

yurkovychv:
1. enabled custom locator plugin
2. updated some selectors 
3. changed xpath selectors to css and locate(uses xpath) selectors
4. added comments to clarify methods


pmm-app/tests/pages/adminPage.js
```diff
@@ -14,16 +14,19 @@ module.exports = {
...
+        .withAttr({'data-toggle' : 'dropdown'})
+        .find('span')
+        .withText('PMM')
 
   },
 
   // introducing methods
   dropdownMenuItemLocator(title) {
-    return (`//ul/li/a[text()='${title}']`)
+    return locate('ul')
```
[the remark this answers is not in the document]
yurkovychv: done, thanks

yurkovychv:
locate() method does not work in all cases, but i added it where applicable

pmm-app/package.json
```diff
@@ -58,7 +58,7 @@
     "@types/react-dom": "^16.9.6",
     "@types/react-router-dom": "^5.1.5",
     "babel-jest": "^24.9.0",
-    "codeceptjs": "^2.6.1",
+    "codeceptjs": "^2.6.5",
```
atymchuk, line 61: When upgrading packages related to the test suite, we should also try to keep companion packages upgraded. For example, we could potentially upgrade:
- codeceptjs-resemblehelper to v1.9.1
- playwright to v1.0.2
- playwright-video to v1.0.0
- codecov v3.7.0

This can be done one by one, if the tests pass - good, if they don't - rollback to the current version.
Actually, there may breaking changes for playwright, so proceed with care :)

pmm-app/pr.codecept.js
```diff
@@ -2,8 +2,9 @@ exports.config = {
   output: 'tests/output',
   helpers: {
     Playwright: {
-      url: 'http://localhost/',
+      url: 'http://18.221.19.47/',
```
atymchuk, line 5: Are we sure we want to commit this change?

pmm-app/tests/pages/adminPage.js
```diff
@@ -14,16 +14,16 @@ module.exports = {
     metricTitle: "//div[@class='panel-title']",
     reportTitleWithNA:
       "//span[contains(text(), 'N/A')]//ancestor::div[contains(@class,'panel-container')]//span[contains(@class,'panel-title-text')]",
-    sttDisabledFailedChecksPanelSelector: "//div[@data-qa='db-check-panel-settings-link']",
-    sttFailedChecksPanelSelector:"//div[@data-qa='db-check-panel-has-checks']",
-    checksPanelSelector:"//div[@data-qa='db-check-panel-home']",
-    pmmDropdownMenuSelector: "//a[@data-toggle='dropdown']/span[text()='PMM']",
+    pmmDropdownMenuSelector: locate('a')
```
atymchuk, line 17: This could possibly be `locate('a[data-toggle="dropdown"] span').withText('PMM')`

pmm-app/tests/pages/adminPage.js
```diff
@@ -14,16 +14,16 @@ module.exports = {
...
+        .withAttr({'data-toggle' : 'dropdown'})
+        .find('span')
+        .withText('PMM')
 
   },
 
   // introducing methods
   dropdownMenuItemLocator(title) {
-    return (`//ul/li/a[text()='${title}']`)
+    return locate('ul > li > a').withText(`${title}`);
```
atymchuk, line 26:
```suggestion
    return locate('ul > li > a').withText(title);
```

atymchuk: This is a variable, you can pass it as is.

pmm-app/tests/pages/homePage.js
```diff
@@ -13,18 +13,18 @@ module.exports = {
...
-    reloadButtonAfterUpgrade: "//button[@ng-click='reloadAfterUpdate()']",
-    pmmUpdateWidget: "//div[@id='pmm-update-widget']",
-    upToDateLocator: "//section[@class='state']/p[text() = 'You are up to date']",
-    availableVersion: "//div[@id='available_version']/div/p",
-    currentVersion: "//p[@id='current_version']/span",
-    lastCheckSelector: "//div[@class='last-check-wrapper']",
-    sttDisabledFailedChecksPanelSelector: "//div[@data-qa='db-check-panel-settings-link']",
-    sttFailedChecksPanelSelector:"//div[@data-qa='db-check-panel-has-checks']",
-    checksPanelSelector:"//div[@data-qa='db-check-panel-home']",
+    checkUpdateButton: "//*[@id='refresh']",
```
atymchuk, line 16:
```suggestion
    checkUpdateButton: "#refresh",
```

atymchuk: This should work.

pmm-app/package.json
atymchuk: Just discovered that playwright 1.x is supported since v2.6.4.

pmm-app/pr.codecept.js
atymchuk: If yes, consider documenting what it is.

pmm-app/tests/pages/databaseChecksPage.js
```diff
@@ -9,36 +11,40 @@ module.exports = {
...
-    tooltipSelector: "//div[@class='ant-tooltip-inner']/div/div[1]",
-    totalFailedChecksTooltipSelector: "//div[@class='popper']//div[text()='Failed checks: ']",
-    failedChecksTooltipSelector: "//div[@class='popper']/div/div/div"
+    dbCheckPanelSelector: "$db-check-panel",
+    dbCheckPanelEmptySelector: "$db-check-panel-table-empty",
+    sttEnabledDBCheckPanelSelector: "$db-check-panel-home",
+    disabledSTTMessageSelector: "$db-check-panel-settings-link",
+    serviceNameSelector: "tbody > tr > td:first-child",
+    totalFailedChecksTooltipSelector: ".popper > div > div > div:first-of-type",
+    failedChecksTooltipSelector: ".popper > div > div > div",
```
puneet0191, line 20: @atymchuk the more I see the CSS selectors, I don't agree with them, atleast I don't see any difference either in the readability of the locators, or them being tightly coupled to HTML structure as we discussed because to me 
`.popper > div > div > div` in no different then `//div[@class='popper']/div/div/div`, It is faster to prepare xpath locators over CSS for sure.

pmm-app/pr.codecept.js
yurkovychv: thats my mistake, i'll fix that , thanks

pmm-app/tests/pages/databaseChecksPage.js
atymchuk: @puneet0191 There are two reasons for this:
1/ css is native to the browser and would work in the browser, so this is something any developer would understand, which is not the case with xpath. You happen to take a very simple selector, but example like this one demonstrate the case much better: `"//span[contains(text(), 'N/A')]//ancestor::div[contains(@class,'panel-container')]//span[contains(@class,'panel-title-text')]"`. We can refer to sites like codecept.io and many others, who also recommend to prefer ccs over xpath. And they do it for the very same reason.
2/ Decoupling the selectors is a real challenge. It's hardly achievable to do it everywhere, but as we discussed, we will try to address it as much as we can by using custom locators and specific `data-` attributes. The readability there is outstanding, because we can give them semantic names.
3/ The time to make an css selector is equal or inferior to the time to make an xpath selector. The modern browsers feature excellent tooling for both of them (see attachment).
4/ If I want to test myself in the browser, even before I launch the test, I will be able to do so using a css selector, which is not the case with an xpath selector. For example `document.querySelectorAll('.popper > div > div > div')` would do the job right from the console.

pmm-app/package.json
yurkovychv: moved version back to 2.6.1, looks like we might need to update it along with updating docker image version

pmm-app/tests/pages/homePage.js
yurkovychv: done thanks

pmm-app/tests/pages/adminPage.js
yurkovychv: done thanks

yurkovychv:
UI tests job passed successfully. Merging PR

yurkovychv:
restored branch, will make some refactoring and create another PR

PR #741, opened 2020-05-27
PMM-6033 fix height on panel wrappers

atymchuk:
- [ ] https://github.com/Percona-Lab/pmm-submodules/pull/869


atymchuk:
@puneet0191 Can we throw away the webdriverio?

puneet0191:
@atymchuk please go ahead, as I know we don't use it anymore on any jobs

PR #760, opened 2020-06-10
SAAS-155 - STT Permissions UI tests

yurkovychv:
[SAAS-155](https://jira.percona.com/browse/SAAS-155)

1. Implemented methods with API calls for creating a user, setting a role and deleting user
2. Added tests for checking correct message when user has not access to the panel/page

### DIFF
--- a/pmm-app/tests/helper/grafana_helper.js
+++ b/pmm-app/tests/helper/grafana_helper.js
@@ -1,5 +1,6 @@
 const Helper = codecept_helper;
-const { saveVideo } = require('playwright-video');
+const assert = require('assert');
+const {saveVideo} = require('playwright-video');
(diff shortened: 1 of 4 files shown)


pmm-app/tests/helper/grafana_helper.js
```diff
@@ -1,5 +1,6 @@
 const Helper = codecept_helper;
-const { saveVideo } = require('playwright-video');
+const assert = require('assert');
+const {saveVideo} = require('playwright-video');
```
puneet0191, line 3: why ?

yurkovychv: these are just formatting changes

PR #771, opened 2020-06-15
PMM-6072 Tests for Upgrade widget, Upgrade tests

puneet0191:
[PMM-6072](https://jira.percona.com/browse/PMM-6072)


pmm-app/tests/upgradePMM_test.js
```diff
@@ -4,92 +4,132 @@ const serviceNames = {
...
+    const dataRetentionActualValue = await I.grabValueFrom(pmmSettingsPage.fields.dataRetentionCount);
+    assert(
+      dataRetention,
+      dataRetentionActualValue,
+      'The Value for Data Retention is not the same as passed via Docker Environment Variable'
+    );
+  }
+);
+
+Scenario(
```
atymchuk, line 96: I find it unusual to verify the presence of the news panel in a test called `upgradePMM_test.js`. Shouldn't each test be responsible for only one specific task? In this case one could expect us to only test whether the upgrade is successful or not, pls some post-upgrade sanity checks.

pmm-app/tests/upgradePMM_test.js
```diff
@@ -4,92 +4,132 @@ const serviceNames = {
...
-        await inventoryAPI.verifyServiceExistsAndHasRunningStatus(service,
-            serviceNames[service.service]);
-      }
+  'PMM-T289 Verify Whats New link is presented on Update Widget @pmm-upgrade @visual-test @not-pr-pipeline',
+  async (I, homePage) => {
+    I.amOnPage(homePage.url);
+    const res2 = process.env.DOCKER_VERSION.split('.');
+    const res = process.env.PMM_SERVER_LATEST.split('.');
+    const majorVersionDiff = res[1] - res2[1];
+    const patchVersionDiff = res[2] - res2[2];
```
atymchuk, line 24:
```suggestion
    const [, pmmMajor, pmmMinor] = process.env.PMM_SERVER_LATEST.split('.');
    const [, dockerMajor, dockerMinor] = process.env.DOCKER_VERSION.split('.');
    const majorVersionDiff = pmmMajor - dockerMajor;
    const patchVersionDiff = pmmMinor - dockerMinor;
```

atymchuk: 1/ This would make it easier to understand what it does.
2/ We would also need to write a comment providing some details to this comparison (i.e. why)

pmm-app/tests/upgradePMM_test.js
```diff
@@ -4,92 +4,132 @@ const serviceNames = {
...
 );
 
 Scenario(
-    'Verify Agents are RUNNING after Upgrade (API) @pmm-upgrade @visual-test @not-pr-pipeline',
-    async (I, inventoryAPI) => {
-      for (const service of Object.values(inventoryAPI.services)) {
-        await inventoryAPI.verifyServiceExistsAndHasRunningStatus(service,
-            serviceNames[service.service]);
-      }
+  'Verify user can create Remote Instances before upgrade and they are in RUNNNING status @pmm-upgrade @visual-test @not-pr-pipeline',
```
atymchuk, line 43: This may be unrelated to testing PMM Upgrade functionality

pmm-app/tests/upgradePMM_test.js
puneet0191: sure makes sense, I have tried added clarification and followed your suggestion.

pmm-app/tests/upgradePMM_test.js
puneet0191: Well for us we define what we are doing in the Feature and for us, this test file Feature is about testing Upgrade and upgrade related test cases from our test cycle. 
`Feature('PMM server Upgrade Tests and Executing test cases related to Upgrade Testing Cycle');` 
I have modified the Feature with more detail to make it clear.

pmm-app/tests/upgradePMM_test.js
puneet0191: yes, this is part of our post-upgrade sanity checks, so we are executing it.

pmm-app/tests/upgradePMM_test.js
atymchuk: Got it, it's clear now with that description :)

PR #824, opened 2020-07-03
PMM-6177-Add UI tests for Delete option in PMM Inventory via UI

IvanaWzientkova:
[PMM-6177](https://jira.percona.com/browse/PMM-6177)


pmm-app/tests/pages/pmmInventoryPage.js
```diff
@@ -16,6 +16,11 @@ module.exports = {
     nodesLinkOld: '//a[contains(text(), "Nodes")]',
     pmmAgentLocator: '//table//td[contains(text(), "PMM Agent")]',
     serviceIdLocatorPrefix: '//table//tr/td[4][contains(text(),"',
+    deleteButton: '//span[contains(text(), "Delete")]',
+    proceedButton: '//span[contains(text(), "Proceed")]',
+    forceModeCheckbox:
+      '//div[@data-qa="form-field-force"]/descendant::span[@class="checkbox-container__checkmark"]',
+    tableCheckbox: '//div[@data-qa="select-row"]',
```
atymchuk, line 23: Pls prefer css wherever possible: `div[data-qa="select-row"]`

pmm-app/tests/pages/pmmInventoryPage.js
```diff
@@ -16,6 +16,11 @@ module.exports = {
     nodesLinkOld: '//a[contains(text(), "Nodes")]',
     pmmAgentLocator: '//table//td[contains(text(), "PMM Agent")]',
     serviceIdLocatorPrefix: '//table//tr/td[4][contains(text(),"',
+    deleteButton: '//span[contains(text(), "Delete")]',
+    proceedButton: '//span[contains(text(), "Proceed")]',
+    forceModeCheckbox:
+      '//div[@data-qa="form-field-force"]/descendant::span[@class="checkbox-container__checkmark"]',
```
atymchuk, line 22: this can be simpler `div[data-qa="form-field-force"] span.checkbox-container__checkmark`

pmm-app/tests/pages/pmmInventoryPage.js
IvanaWzientkova: Codecept has a lot of issues with css. Sometimes they don't work, that's why we are back with Xpath. Xpaths are more stable for codecept.

pmm-app/tests/pages/pmmInventoryPage.js
IvanaWzientkova: Same as above :)

pmm-app/tests/pages/pmmInventoryPage.js
atymchuk: this ^ just works in other places, therefore should work here

pmm-app/tests/pages/pmmInventoryPage.js
atymchuk: Our decision was to use xpath only when we can't use css, IIRC

pmm-app/tests/pages/pmmInventoryPage.js
atymchuk: Our decision was to use xpath only when we can't use css, IIRC

IvanaWzientkova: We had the conversation on June 22nd. Look at the screenshot:
image
Look at the code and see that it's failing. Nobody knows why, but it happening sometimes. :)

atymchuk: Looking at the the screenshot both are passing. What am I missing?

PR #840, opened 2020-07-10
PMM-6241 Fix filters percentage format

tiagomotasantos:
[PMM-6241](https://jira.percona.com/browse/PMM-6241)

Adds missing percentage sign to humanize transform when name is percent and uses percentRounded for QAN filters.
Also changes percentRounded to show <1% instead of <0.01%. Since the percentage is rounded there's no need to show decimals.


pmm-app/src/pmm-qan/panel/components/Filters/components/CheckboxGroup/CheckboxGroup.tsx
```diff
@@ -56,7 +56,7 @@ export const CheckboxGroup: FC<CheckboxGroupProps> = ({
             />
           </span>
           <span className={styles.percentage}>
-            <span>{valueExists ? humanize.transform(item.main_metric_percent, 'percent') : null}</span>
+            <span>{valueExists ? humanize.transform(item.main_metric_percent, 'percentRounded') : null}</span>
```
lunaticusgreen, line 59: It shouldn't be rounded here because percentage values may be relly small and we need to differentiate it somehow

pmm-app/src/shared/components/helpers/Humanization.ts
```diff
@@ -81,8 +81,8 @@ export const humanize = {
 
         break;
       case 'percentRounded':
-        if (metricValue !== 0 && metricValue < 0.0001) {
-          res = '<0.01';
+        if (metricValue !== 0 && metricValue < 0.01) {
```
lunaticusgreen, line 84: 0.0001 is also correct value here, because we often will meet values <1%

PR #856, opened 2020-07-16
PMM-6263 refactored QAN tests

IvanaWzientkova:
[PMM-6263](https://jira.percona.com/browse/PMM-6263)


pmm-app/tests/QAN/filters_test.js
```diff
@@ -1,127 +1,222 @@
...
+    I.seeAttributesOnElements(qanPage.fields.previousPage, { 'aria-disabled': 'true' });
+    I.click(qanPage.fields.ellipsisButton);
+    qanActions.verifyActiveItem(6);
+    await qanActions.verifyCount('126-150');
+    I.click(qanPage.fields.ellipsisButton);
+    qanActions.verifyActiveItem(1);
+    await qanActions.verifyCount('1-25');
+    qanActions.selectPage(3);
+    qanActions.verifyActiveItem(3);
+    await qanActions.verifyCount('51-75');
```
lunaticusgreen, line 180: I guess it won't work properly, because we're not checking that we have enough elements ( 75 )

pmm-app/pr.codecept.js
```diff
@@ -38,7 +38,8 @@ exports.config = {
     homePage: './tests/pages/homePage.js',
     remoteInstancesPage: './tests/pages/remoteInstancesPage.js',
     adminPage: './tests/pages/adminPage.js',
-    qanPage: './tests/QAN/qanPage.js',
+    qanPage: './tests/QAN/pages/qanPage.js',
+    qanActions: './tests/QAN/steps/qanActions.js',
```
yurkovychv, line 42: I see the separate Actions page. This is great, but didn't we want to use page fragments? to have separate object for each Component?

puneet0191: technically it is the same page object, just that the scope for this task was to break actions and Locators separate and get these tests up and running, we can have another task for page fragments for QAN page.

pmm-app/tests/QAN/filters_test.js
puneet0191: you are right, I realized it while handling other cases with 25, 50 checks, it worked now because we have a different job for QAN data to be loaded with enough wait to load those distinct queries from all instances hence it worked, but to handle this check completely it would need more "logic".

pmm-app/pr.codecept.js
yurkovychv: oh, thats clear now for me

pmm-app/tests/QAN/pages/qanPage.js
```diff
@@ -0,0 +1,136 @@
...
+    nextPageNavigation: '//ul[@role=\'navigation\']//li[last()]',
+    iframe: '//div[@class=\'panel-content\']//iframe',
+    newQANPanelContent: '.panel-content',
+    disabledResetAll: '//button[@data-qa="qan-filters-reset-all" and @disabled ]',
+    newQANSpinnerLocator: '//i[@data-qa=\'loading-spinner\']',
+    showSelected: '//button[@data-qa=\'qan-filters-show-selected\']',
+    countOfItems: '//span[@data-qa=\'qan-total-items\']',
+    filterBy: '//input[@data-qa=\'filters-search-field\']',
+    filterCheckboxes: '.checkbox-container__checkmark',
+    newQANAddColumn: '//span[contains(text(), \'Add column\')]',
```
atymchuk, line 28: I suggest we use double quotes in such cases. This'd read better than escaping.
```suggestion
    newQANAddColumn: '//span[contains(text(), "Add column")]',
```

pmm-app/tests/QAN/pages/qanPage.js
```diff
@@ -0,0 +1,136 @@
...
+    searchFieldForColumn: 'input.ant-select-search__field',
+    newQANMetricDropDown: '.ant-select-dropdown-menu-item',
+    groupBySelector: '.group-by-selector',
+    addColumnSelector: '.add-columns-selector',
+    removeColumnButton: '//div[text()=\'Remove column\']',
```
atymchuk, line 33:
```suggestion
    removeColumnButton: '//div[text()="Remove column"]',
```

pmm-app/tests/QAN/pages/qanPage.js
atymchuk: Besides, it seems to be a general practice out there.

pmm-app/tests/QAN/pages/qanPage.js
```diff
@@ -0,0 +1,136 @@
...
+    resultsPerPageValue: '.ant-select-selection-selected-value',
+    nextPage: '.ant-pagination-next',
+    previousPage: '.ant-pagination-prev',
+    ellipsisButton: '.ant-pagination-item-ellipsis',
+    tableRow: 'div.tr',
+    resultPerPageCombobox: '.ant-pagination-options',
+    addColumnNewQAN: '//span[contains(text(), \'Add column\')]',
```
atymchuk, line 40:
```suggestion
    addColumnNewQAN: '//span[contains(text(), "Add column")]',
```

atymchuk: Actually, I won't comment each of them, but pls consider this approach and maybe apply it in case it works.

pmm-app/tests/QAN/timerange_test.js
```diff
@@ -6,44 +6,46 @@ Before((I, qanPage) => {
...
     const TIME_RANGE_QUERY_PARAMS_BEFORE = 'from=now-5m&to=now';
     const TIME_RANGE_QUERY_PARAMS_AFTER = 'from=now-3h&to=now';
-    qanPage.waitForNewQANPageLoaded();
+
+    qanActions.waitForNewQANPageLoaded();
     I.seeInCurrentUrl(TIME_RANGE_QUERY_PARAMS_BEFORE);
     adminPage.applyTimeRange('Last 3 hours');
-    //qanPage.waitForResponsePath(qanPage.requests.getReportPath);
-    qanPage.waitForNewQANPageLoaded();
+    // qanPage.waitForResponsePath(qanPage.requests.getReportPath);
```
atymchuk, line 28: Shall we clean it up?

pmm-app/tests/QAN/pages/qanPage.js
puneet0191: will fix. thanks

pmm-app/tests/QAN/timerange_test.js
puneet0191: yeap, will do

PR #874, opened 2020-07-24
PMM-6247 hotfix

lunaticusgreen:
[PMM-6247](https://jira.percona.com/browse/PMM-6247)

By mistake I removed import of this hook and it disabled original task solution, just adding hook back


lunaticusgreen:
@tiagomotasantos PTAL

PR #875, opened 2020-07-24
PMM-6277 Add shared components InputLabel and NumericInput

nicolalamacchia:
[PMM-6277](https://jira.percona.com/browse/PMM-6277)


pmm-app/src/shared/components/Form/NumericInput/NumericInput.test.tsx
```diff
@@ -0,0 +1,108 @@
...
+    (wrapper.find('input').instance() as any).stepUp = mockedStepUp;
+    (wrapper.find('input').instance() as any).stepDown = mockedStepDown;
+
+    expect(mockedStepUp).toBeCalledTimes(0);
+    expect(mockedStepDown).toBeCalledTimes(0);
+
+    wrapper.find('button').at(0).simulate('click');
+
+    expect(mockedStepUp).toBeCalledTimes(1);
+    expect(mockedStepUp).toBeCalledTimes(1);
```
tiagomotasantos, line 72: Duplicate line.

pmm-app/src/shared/components/Form/NumericInput/NumericInput.style.ts
```diff
@@ -0,0 +1,101 @@
+import { css } from 'emotion';
+// import { GrafanaTheme } from '@grafana/data'
+import { stylesFactory } from '@grafana/ui';
+
+// export const getStyles = stylesFactory((theme: GrafanaTheme) => {
+export const getStyles = stylesFactory((theme) => {
```
tiagomotasantos, line 6: I think you can keep the GrafanaTheme type here and remove the comments. If it gives you an error because it doesn't find the colors just do `const { colors }: any = theme;` and use the theme type for other variables (e.g. spacing).

nicolalamacchia: Good catch! Actually it's not giving any errors, it was just a bad copy-paste from the JS sandbox to TS :)

pmm-app/src/shared/components/Form/NumericInput/NumericInput.test.tsx
nicolalamacchia: Fixed

pmm-app/src/shared/components/Form/NumericInput/NumericInput.tsx
```diff
@@ -0,0 +1,83 @@
...
+          </InputLabel>
+        </div>
+      )}
+      <span className={styles.inputWrapper}>
+        <BaseInput
+          {...props}
+          ref={inputRef}
+          type="number"
+          disabled={disabled}
+          className={cx(
```
tiagomotasantos, line 55: Would it make sense to add this class on the root span? Otherwise how to I add a margin to the whole component (label + input)?

tiagomotasantos: Also adding the class to the root makes it possible for the user of the component to style the label

pmm-app/src/shared/components/Form/InputLabel/InputLabel.style.ts
```diff
@@ -0,0 +1,18 @@
+import { css } from 'emotion';
+import { GrafanaTheme } from '@grafana/data';
+import { stylesFactory } from '@grafana/ui';
+
+export const getStyles = stylesFactory((theme: GrafanaTheme) => ({
+  inputLabel: css`
+      align-items: center;
+      background-color: ${theme.colors.bg3};
+      border-radius: 3px 0 0 3px;
```
atymchuk, line 9: @nicolalamacchia Isn't there a borderRadius that can be taken from the theme?

pmm-app/src/shared/components/Form/InputLabel/InputLabel.tsx
```diff
@@ -0,0 +1,15 @@
+import React, { FC, forwardRef } from 'react';
+import { cx } from 'emotion';
+import { useTheme } from '@grafana/ui';
+import { getStyles } from './InputLabel.style';
+
+type InputLabelProps = React.LabelHTMLAttributes<HTMLLabelElement>;
+
+export const InputLabel: FC<InputLabelProps> = forwardRef<HTMLLabelElement, InputLabelProps>(({ children, className = '', ...props }, ref) => {
```
atymchuk, line 8: You can leave className without the default - `undefined` would work as well.

pmm-app/src/shared/components/Form/NumericInput/NumericInput.style.ts
```diff
@@ -0,0 +1,100 @@
+import { css } from 'emotion';
+import { GrafanaTheme } from '@grafana/data';
+import { stylesFactory } from '@grafana/ui';
+
+export const getStyles = stylesFactory((theme: GrafanaTheme) => {
+  const arrow = css`
```
atymchuk, line 6: For the speed sake it's better to put constants which do not depend on anything outside of the function.

pmm-app/src/shared/components/Form/NumericInput/NumericInput.style.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    input {
+      box-sizing: border-box;
+      border-radius: 3px;
+      padding-right: 1.4em;
+    }
+
+    &,
+    &:hover {
+      input[type='number']::-webkit-outer-spin-button,
```
atymchuk, line 46: I wonder what will it be like in non-webkit browsers?

pmm-app/src/shared/components/Form/NumericInput/NumericInput.tsx
```diff
@@ -0,0 +1,83 @@
...
+  const stepDown = () => {
+    inputRef.current!.stepDown();
+    dispatchChangeEvent();
+  };
+
+  return (
+    <span className={styles.wrapper}>
+      {label && (
+        <div>
+          <InputLabel style={labelWidth ? { width: labelWidth * 8, minWidth: 0 } : undefined}>
```
atymchuk, line 44: Not sure if we want to have it as a factor of eight. Seems to be less flexible than just passing an arbitrary number.

pmm-app/src/shared/components/Form/NumericInput/NumericInput.style.ts
nicolalamacchia: AFAIK other browsers always show spin buttons (BTW, this is the approach I found the the Input component by Grafana)

pmm-app/src/shared/components/Form/NumericInput/NumericInput.tsx
nicolalamacchia: Yes, as we discussed: this is the approach Grafana's taking for their components, we're free to use what we want. Just notice that it will behave differently from other native Grafana components (which use multiples of 8) if we use arbitrary numbers

atymchuk: Ok, let's leave 8x for the time being. I guess we need to see it practice then maybe revisit.

pmm-app/src/shared/components/Form/NumericInput/NumericInput.style.ts
nicolalamacchia: Right, moving `arrow` and `baseInput` outside

pmm-app/src/shared/components/Form/InputLabel/InputLabel.tsx
nicolalamacchia: Right, done

pmm-app/src/shared/components/Form/InputLabel/InputLabel.style.ts
nicolalamacchia: You're right, there is! Updated

pmm-app/src/shared/components/Form/NumericInput/NumericInput.tsx
nicolalamacchia: Definitely, my mistake, updating

PR #882, opened 2020-07-28
PMM-6370 add eslint command for e2e files

atymchuk:
[PMM-6370](https://jira.percona.com/browse/PMM-6370)

@yurkovychv Should I let you fix the rest :) or maybe we could team up and fix the rest together?


yurkovychv:
However you say @atymchuk :) The thing is that i'm not experienced in linters tuning, so i'll definitely need your help with this

puneet0191:
thanks a lot for this :+1:

puneet0191:
just need to check job once, looks great

PR #889, opened 2020-07-30
PMM-6272 refactored PMM settings page

atymchuk:
[PMM-6272](https://jira.percona.com/browse/PMM-6272)

Branched restored after an accidental merge.
FB: https://github.com/Percona-Lab/pmm-submodules/pull/992


pmm-app/src/pmm-settings/Settings.tsx
```diff
@@ -0,0 +1,119 @@
...
+  const updateSettings = async (body: any, callback: LoadingCallback) => {
+    const response = await SettingsService.setSettings(body, callback);
+
+    if (response) {
+      setSettings(response);
+    }
+  };
+
+  useEffect(() => {
+    SettingsService.getSettings(setLoading, setSettings);
```
nicolalamacchia, line 56: I don't really like this approach, is there a particular reason for using callbacks here? I find this a bit obscure.

pmm-app/src/pmm-settings/Settings.service.ts
```diff
@@ -1,10 +1,53 @@
...
+      console.error(e);
+    } finally {
+      setLoading(false);
+    }
+
+    return response;
   },
 };
+
+const toModel = (response: any): Settings => ({
```
nicolalamacchia, line 43: why not adding an interface for `response` instead of using `any` here?

pmm-app/src/pmm-settings/Settings.service.ts
```diff
@@ -1,10 +1,53 @@
...
+    } finally {
+      setLoading(false);
+    }
+
+    return response;
   },
-  setSettings(body) {
-    return apiRequest.post<any, any>('/v1/Settings/Change', body);
+  async setSettings(body, setLoading: LoadingCallback) {
+    let response: any;
```
nicolalamacchia, line 26: I think the structure of `response` should be known: there should be a more specific type here instead of `any`

pmm-app/src/pmm-settings/Settings.service.ts
```diff
@@ -1,10 +1,53 @@
...
+
+    try {
+      setLoading(true);
+      response = await apiRequest.post<any, any>('/v1/Settings/Get', {});
+      setSettings(toModel(response.settings));
+    } catch (e) {
+      console.error(e);
+    } finally {
+      setLoading(false);
```
nicolalamacchia, line 20: I would use `setSettings` and `setLoading` outside this function and use `getSettings` just to fetch data

pmm-app/src/pmm-settings/Settings.tsx
tiagomotasantos: Yes, the callbacks were not to block the main execution waiting for the request and don't have to bring those try/catch blocks here. 
I was using the `SettingsService.setSettings` on each separated component at first, and having to replicate the same error handling code and loading was too much duplication. And the API layer is not dependent on the model because it defines the the callback interfaces that it expects (so it depends on an abstraction and not concrete objects).
I saw the other comments and yes I agree that the presentation layer shouldn't know nothing about the requests, maybe by having a middle layer to transform the model to API or vice versa. Or use any external library, but we definitely have to work on that.

PR #898, opened 2020-08-06
PMM-6372 sign up page

nicolalamacchia:
[PMM-6372](https://jira.percona.com/browse/PMM-6372)

https://github.com/Percona-Lab/pmm-submodules/pull/1013


pmm-app/src/pmm-settings/components/SignUp/SignUp.tsx
```diff
@@ -0,0 +1,126 @@
...
+import { Messages } from './SignUp.messages';
+import { InputFieldAdapter, CheckboxFieldAdapter } from './FieldAdapters/FieldAdapters';
+import { TERMS_OF_SERVICE_URL, PRIVACY_POLICY_URL, NOTIFICATION_SETTINGS_URL } from './SignUp.constants';
+import { required, validEmail, composeValidators } from './SignUp.validators';
+import { SignUpData, SignUpProps } from './types';
+import { LoggedIn } from './LoggedIn/LoggedIn';
+
+const signUp = async (signUpData: SignUpData) => {
+  // XXX: stub for the actual sign up API call
+  signUpData;
```
atymchuk, line 23: If it doesn't do anything, let's comment it out.

pmm-app/src/pmm-settings/components/SignUp/SignUp.tsx
```diff
@@ -0,0 +1,126 @@
...
+  await new Promise((resolve) => setTimeout(resolve, 1500));
+};
+
+export const SignUp: FC<SignUpProps> = ({ userEmail }) => {
+  const theme = useTheme();
+  const styles = getStyles(theme);
+  const [loggedInEmail, setLoggedInEmail] = useState(userEmail);
+
+  const checkboxLabel = (
```
atymchuk, line 32: It's better to respect the component naming convention, i.e. CheckboLabel

atymchuk: Even though it's internal to to the parent component.

pmm-app/src/pmm-settings/components/SignUp/SignUp.validators.ts
```diff
@@ -0,0 +1,20 @@
+import { Messages } from './SignUp.messages';
+
+export const required = (value: string | boolean | undefined) => (
+  value ? undefined : Messages.errors.requiredField
```
atymchuk, line 4: A boolean false would give a false positive. Consider `value != null` (mind the single equal sign).

pmm-app/src/pmm-settings/components/SignUp/SignUp.validators.ts
```diff
@@ -0,0 +1,20 @@
...
+);
+
+export const validEmail = (value: string) => {
+  const emailRe = RegExp(''
```
atymchuk, line 8: Let's try to get away with regex literal using forward slashes only. The constructor wouldn't be that justified here )

atymchuk: eslint should be have, or let teach it to be )

pmm-app/src/pmm-settings/components/SignUp/SignUp.tsx
nicolalamacchia: this is to avoid to TS complaining about unused variable

pmm-app/src/pmm-settings/components/SignUp/SignUp.tsx
nicolalamacchia: the reason I didn't follow that naming convention is because that's not actually a component, but a piece of JSX code. In this case it would be even better to convert it to a component and pass it as `label={CheckboxLabel}`.

pmm-app/src/pmm-settings/components/SignUp/SignUp.validators.ts
nicolalamacchia: I agree, this was an attempt to avoid having a really long line, since it's not practical to break a regex literal on several lines. Do you have an advice for that?

pmm-app/src/pmm-settings/components/SignUp/SignUp.tsx
nicolalamacchia: Done

pmm-app/src/pmm-settings/components/SignUp/SignUp.validators.ts
atymchuk: yes, it should just be a regex, although a bit long )

pmm-app/src/pmm-settings/components/SignUp/SignUp.validators.ts
nicolalamacchia: In this case `false ? undefined : Messages.errors.requiredField` I want it to return the error message actually, since the value should be `true` in order to make it pass (think about the checkbox that must be checked). Even splitting that to two different validators, like `requireTrue` and `required` or something like that, the logic for both would be the same. Well, actually it could be more specific for the hypothetical `requireTrue` validator, the condition could be stricter: `value === true`, but I'm not sure it's worth the effort. What do you think?

pmm-app/src/pmm-settings/components/SignUp/SignUp.tsx
```diff
@@ -0,0 +1,126 @@
...
+  LinkButton,
+  useTheme,
+} from '@grafana/ui';
+import { showSuccessNotification, showErrorNotification } from 'shared/components/helpers';
+import { ButtonWithSpinner } from 'shared/components/Form';
+import { getStyles } from './SignUp.styles';
+import { Messages } from './SignUp.messages';
+import { InputFieldAdapter, CheckboxFieldAdapter } from './FieldAdapters/FieldAdapters';
+import { TERMS_OF_SERVICE_URL, PRIVACY_POLICY_URL, NOTIFICATION_SETTINGS_URL } from './SignUp.constants';
+import { required, validEmail, composeValidators } from './SignUp.validators';
```
lunaticusgreen, line 17: I guess that this functionality should be moved to shared, we already have something for validation there

nicolalamacchia: good idea!

pmm-app/src/pmm-settings/components/SignUp/SignUp.validators.ts
nicolalamacchia: Done, eslint's `max-len` will ignore regex literals

pmm-app/src/pmm-settings/components/SignUp/SignUp.tsx
nicolalamacchia: Done

pmm-app/src/pmm-settings/components/SignUp/FieldAdapters/FieldAdapters.styles.ts
```diff
@@ -0,0 +1,13 @@
...
+import { GrafanaTheme } from '@grafana/data';
+import { stylesFactory } from '@grafana/ui';
+
+export const getStyles = stylesFactory((theme: GrafanaTheme) => ({
+  errorMessage: css`
+    color: ${(theme.colors as any).red};
+    font-size: 10px;
+    height: 10px;
+    line-height: 1;
+    margin-top: 3px;
```
tiagomotasantos, line 11: Use theme values if you think is right instead of hardcoded values (e.g. theme.spacing.sm) and in other files as well

pmm-app/src/pmm-settings/components/SignUp/SignUp.tsx
```diff
@@ -0,0 +1,126 @@
...
+    pristine, submitting, valid, handleSubmit
+  }) => (
+    <form data-qa="sign-up-form" className={styles.form} onSubmit={handleSubmit}>
+      <legend className={styles.legend}>
+        {Messages.signUp}
+      </legend>
+      <FinalField
+        data-qa="sign-up-email-input"
+        name="email"
+        label="Email"
```
tiagomotasantos, line 69: Should these labels be in the messages file?

pmm-app/src/pmm-settings/components/SignUp/FieldAdapters/FieldAdapters.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+export const InputFieldAdapter = ({
+  input, className, label, meta, ...props
+}) => {
+  const theme = useTheme();
+  const styles = getStyles(theme);
+
+  return (
+    <Field label={label}>
+      <>
```
tiagomotasantos, line 20: I think if you set the children prop in the FieldProps to ReactNode instead of ReactElement you don't need the Fragment. Anyway I understand that the Field is just temporary. And I wonder once we migrate to Grafana 7 Field component if it's better to put these Adapters on the shared or refactor the existing ones in there.

pmm-app/src/pmm-settings/components/SignUp/SignUp.tsx
nicolalamacchia: Good point, done

pmm-app/src/pmm-settings/components/SignUp/FieldAdapters/FieldAdapters.styles.ts
nicolalamacchia: Good point, done

pmm-app/src/pmm-settings/components/SignUp/FieldAdapters/FieldAdapters.tsx
nicolalamacchia: That's because the `Field` components comes directly from Grafana. I moved it temporarily here in order to avoid conflicts between different Grafana versions. It will disappear once we migrate to Grafana 7, for this reason I prefer to leave it as it is for now :)

pmm-app/src/pmm-settings/components/SignUp/SignUp.validators.ts
atymchuk: I think in case of a boolean value we expect the input to validate in either case: true or false. Null or undefined only should trigger the validation error. And this is not the same as saying `(value: boolean) => value ? undefined : Message;`

pmm-app/src/pmm-settings/components/SignUp/SignUp.tsx
```diff
@@ -0,0 +1,126 @@
+import React, {
+  FC,
+  useState
+} from 'react';
+import { Form as FinalForm, Field as FinalField, FormRenderProps } from 'react-final-form';
```
atymchuk, line 5: I would probably not rename the exports unless there is a clash between those and some other names.

pmm-app/src/pmm-settings/components/SignUp/SignUp.tsx
```diff
@@ -0,0 +1,126 @@
...
+      >
+        {Messages.signIn}
+      </Button>
+    </form>
+  );
+
+  return (
+    // TODO (nicolalamacchia): improve this once the sign in component is ready
+    <>
+      {typeof loggedInEmail !== 'undefined'
```
atymchuk, line 114: I guess a much better approach would be to just check the value.
```suggestion
      {loggedInEmail != null
```

pmm-app/src/pmm-settings/components/SignUp/SignUp.tsx
```diff
@@ -0,0 +1,126 @@
...
+  const theme = useTheme();
+  const styles = getStyles(theme);
+  const [loggedInEmail, setLoggedInEmail] = useState(userEmail);
+
+  const CheckboxLabel = () => (
+    <span data-qa="sign-up-agreement-checkbox-label" className={styles.checkboxLabel}>
+      {Messages.agreementFirstPart}
+      {' '}
+      <LinkButton className={styles.link} variant="link" href={TERMS_OF_SERVICE_URL}>{Messages.termsOfService}</LinkButton>
+      ,
```
atymchuk, line 37: Let's put the comma inside of the curlies, i.e. `{', '}`

pmm-app/src/pmm-settings/components/SignUp/SignUp.tsx
```diff
@@ -0,0 +1,126 @@
...
+      {' '}
+      <LinkButton className={styles.link} variant="link" href={PRIVACY_POLICY_URL}>{Messages.privacyPolicy}</LinkButton>
+      ,
+      {' '}
```
atymchuk, line 41: ditto

pmm-app/src/pmm-settings/components/SignUp/SignUp.tsx
nicolalamacchia: You're right! There was a clash before, but not anymore :) Done

pmm-app/src/pmm-settings/components/SignUp/SignUp.tsx
nicolalamacchia: Good point, done

pmm-app/src/pmm-settings/components/SignUp/SignUp.tsx
nicolalamacchia: done

pmm-app/src/pmm-settings/components/SignUp/SignUp.tsx
nicolalamacchia: Done

pmm-app/src/pmm-settings/components/SignUp/SignUp.validators.ts
nicolalamacchia: Got your point, using a `requiredTrue` validator for that now

PR #909, opened 2020-08-12
PMM-6475 add parameter hooks

lunaticusgreen:
[PMM-6475](https://jira.percona.com/browse/PMM-6475)

https://github.com/Percona-Lab/pmm-submodules/pull/1027


atymchuk:
@lunaticusgreen Can we have a FB please?

pmm-app/src/shared/components/hooks/parameters.hook.ts
```diff
@@ -0,0 +1,26 @@
+import { useEffect, useState } from 'react';
+import { getLocationSrv } from '@grafana/runtime';
+
+export const useUrlState = (key: string, defaultValue?: string): [string, (parameter: string) => void] => {
```
atymchuk, line 4: This name is too generic. A more semantic name would be `useQueryParams`. 
@tiagomotasantos @nicolalamacchia Unless there is a better name :)

lunaticusgreen:
> @lunaticusgreen Can we have a FB please?

Done

pmm-app/src/shared/components/hooks/parameters.hook.ts
```diff
@@ -0,0 +1,26 @@
...
+  const query = new URLSearchParams(window.location.search);
+  const [value, setValue] = useState<string>(defaultValue || '');
```
nicolalamacchia, line 6: shouldn't this be `??` instead of `||`?

pmm-app/src/shared/components/hooks/parameters.hook.ts
nicolalamacchia, line 4: I think it would be better to separate the typespec here

pmm-app/src/shared/components/hooks/parameters.hook.ts
atymchuk: I should. It's our muscle memory :)

tiagomotasantos (approved):

atymchuk (approved):

PR #913, opened 2020-08-13
Fix UI Tests

IvanaBrecherova:


pmm-app/tests/QAN/filters_test.js
```diff
@@ -269,7 +269,7 @@ Scenario(
     I.waitForElement(qanPage.fields.noQueries, 20);
   },
 );
-
+*/
```
puneet0191, line 272: @IvanaBrecherova  I am not sure why comment?

pmm-app/tests/QAN/pages/qanPage.js
```diff
@@ -65,6 +65,7 @@ module.exports = {
     table: '//table//tr[2]',
     tableRow: 'div.tr',
     noQueries: "//h1[contains(text(), 'No queries available for this combination of filters')]",
+    Last12HoursValue: "//span[contains(text(), 'Last 12 hours')]",
```
puneet0191, line 68: This should already be available in DashboardPage, if not, thats the right place for such locator.

pmm-app/tests/QAN/steps/qanActions.js
```diff
@@ -175,7 +175,7 @@ module.exports = {
   },
 
   verifyChangedCount(countBefore, countAfter) {
-    assert.notEqual(countAfter, countBefore, 'Data should be changed');
+    assert.notEqual(countAfter, countBefore, `The value ${countBefore} should not be equal to ${countAfter}`);
```
puneet0191, line 178: Please add extra Parameter for the type of verification, so in case you are checking results after applying filter, it should say `After applying Filter value ${countBefore} should not be equal to ${countAfter}` 

Another example: when changing Number of Items, by changing items per page, in that case it should says 
`After applying Item Change value ${countBefore} should not be equal to ${countAfter}`

pmm-app/tests/QAN/filters_test.js
IvanaBrecherova: Forgot to delete, thanks!

IvanaBrecherova:
@puneet0191 Tests passed: https://pmm.cd.percona.com/blue/organizations/jenkins/pmm2-ui-test-nightly/detail/pmm2-ui-test-nightly/79/pipeline 👍

puneet0191 (approved):

PR #918, opened 2020-08-19
PMM-6373 sign in

lunaticusgreen:
[PMM-6373](https://jira.percona.com/browse/PMM-6373)


pmm-app/src/pmm-settings/components/PlatformLogin/SignIn.tsx
```diff
@@ -0,0 +1,83 @@
...
+import { Field, Form, FormRenderProps } from 'react-final-form';
+import { Button, useTheme } from '@grafana/ui';
+import { Credentials, LoginFormProps } from './types';
+import { Messages } from './PlatformLogin.messages';
+import { InputFieldAdapter } from './FieldAdapters/FieldAdapters';
+import validators from '../../../shared/components/helpers/validators';
+import { ButtonWithSpinner } from '../../../shared/components/Form';
+import { getStyles } from './PlatformLogin.styles';
+import { PlatformLoginService } from './PlatformLogin.service';
+import { showErrorNotification, showSuccessNotification } from '../../../shared/components/helpers';
```
nicolalamacchia, line 11: Please, use absolute imports

nicolalamacchia (changes requested):
I noticed there's a lot of boilerplate and duplicated code in `SignIn` and SignUp`

pmm-app/src/pmm-settings/components/PlatformLogin/SignIn.tsx
```diff
@@ -0,0 +1,83 @@
...
+      showErrorNotification({ message: Messages.errors.signUpFailed });
+    }
+  };
+
+  const SignInForm: FC<FormRenderProps<Credentials>> = ({
+    pristine,
+    submitting,
+    valid,
+    handleSubmit,
+    form,
```
nicolalamacchia, line 38: `form` is unused in this file

pmm-app/src/pmm-settings/components/PlatformLogin/SignIn.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+export const SignIn = ({ changeMode, getSettings }: LoginFormProps) => {
+  const theme = useTheme();
+  const styles = getStyles(theme);
+
+  const handleSignInFormSubmit = async (credentials: Credentials) => {
+    try {
+      await PlatformLoginService.signIn(credentials);
+
+      getSettings();
```
nicolalamacchia, line 22: It's not clear to me why we have to call `getSettings` in these pages. also there's a bit of prop drilling to pass the callback here.

pmm-app/src/pmm-settings/components/PlatformLogin/SignUp.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+export const SignUp = ({ changeMode, getSettings }: LoginFormProps) => {
+  const theme = useTheme();
+  const styles = getStyles(theme);
+
+  const handleSignUpFormSubmit = async (credentials: Credentials) => {
+    try {
+      await PlatformLoginService.signUp(credentials);
+
+      getSettings();
```
nicolalamacchia, line 22: See previous comment about `getSettings`

pmm-app/src/pmm-settings/components/PlatformLogin/SignUp.tsx
```diff
@@ -0,0 +1,111 @@
...
+      />
+      <Field
+        data-qa="sign-up-password-input"
+        name="confirm-password"
+        label={Messages.confirmPasswordLabel}
+        type="password"
+        component={InputFieldAdapter}
+        validate={validators.matches('password', Messages.errors.passwordMismatch)}
+        autoComplete="on"
+      />
```
nicolalamacchia, line 78: this has the same `data-qa` as the previous one. also, was it requested to have a password confirmation here?

pmm-app/src/pmm-settings/components/PlatformLogin/SignUp.tsx
```diff
@@ -0,0 +1,111 @@
...
+      {`${Messages.agreementFirstPart} `}
+      <LinkButton className={styles.link} variant="link" href={TERMS_OF_SERVICE_URL}>
+        {Messages.termsOfService}
+      </LinkButton>
+      {` ${Messages.agreementSecondPart} `}
+      <LinkButton className={styles.link} variant="link" href={PRIVACY_POLICY_URL}>
+        {Messages.privacyPolicy}
+      </LinkButton>
+    </span>
+  );
```
nicolalamacchia, line 41: this could be brought outside with the checkbox field (creating a separate component), in order to simplify this code

pmm-app/src/pmm-settings/components/PlatformLogin/PlatformLogin.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { SignUpProps } from './types';
+import { LoggedIn } from './LoggedIn/LoggedIn';
+import { SignIn } from './SignIn';
+import { SignUp } from './SignUp';
+
+export const PlatformLogin: FC<SignUpProps> = ({ userEmail, getSettings }) => {
+  const theme = useTheme();
+  const styles = getStyles(theme);
+  const [isSignInMode, setMode] = useState(true);
+  const toggleMode = () => setMode(!isSignInMode);
```
nicolalamacchia, line 13: this should be `setMode((currentMode) => !currentMode)`. doing this you're not dependent on `isSignInMode` in `setMode`

nicolalamacchia (changes requested):
I noticed there's a lot of boilerplate and duplicated code in `SignIn` and `SignUp`

pmm-app/src/pmm-settings/components/PlatformLogin/SignUp.tsx
lunaticusgreen: I'll change data qa, thanks! No, it wasn't, I just added it because I guess that it really makes sense

pmm-app/src/pmm-settings/components/PlatformLogin/SignIn.tsx
lunaticusgreen: we need to call getSettings because we're not getting email from the backend anymore in answer, so after we sign in we need to re-request data to display login state

pmm-app/src/pmm-settings/components/PlatformLogin/SignUp.tsx
lunaticusgreen: The same as for previous, I'm not sure if we have more elegant solution here. @nicolalamacchia maybe you have any ideas? if yes, please feel free to add it

pmm-app/src/pmm-settings/components/PlatformLogin/SignUp.tsx
lunaticusgreen: here i am afraid it will break the uniformity,  isn'it easier to read it when we have only `Field` components everywhere in this form?

lunaticusgreen:
> I noticed there's a lot of boilerplate and duplicated code in `SignIn` and `SignUp`

pmm-app/src/pmm-settings/components/PlatformLogin/SignUp.tsx
nicolalamacchia: I'd discuss about this also with @atymchuk, since I think the trend now is to have just one password field for sign ups, with no confirmation, but with a "restore password" link.

pmm-app/src/pmm-settings/components/PlatformLogin/SignUp.tsx
nicolalamacchia: I was thinking about moving out of these components the logic and have `signIn` and `singUp` functions in the `PlatformLogin` component, having those components as "dumb" components, in order to avoid that prop drilling and maybe also share more code

pmm-app/src/pmm-settings/components/PlatformLogin/SignUp.tsx
lunaticusgreen: Good point, lets wait for @atymchuk comments, this functionality could be removed few minutes if we need it

pmm-app/src/pmm-settings/components/PlatformLogin/SignIn.tsx
```diff
@@ -0,0 +1,79 @@
...
+      showSuccessNotification({ message: `${Messages.loginSucceeded} ${credentials.email}` });
+    } catch (e) {
+      console.error(e);
+      showErrorNotification({ message: Messages.errors.signUpFailed });
```
nicolalamacchia, line 26: Shouldn't this be a different message from sign up?

pmm-app/src/pmm-settings/components/PlatformLogin/SignIn.tsx
```diff
@@ -0,0 +1,79 @@
...
+export const SignIn = ({ changeMode, getSettings }: LoginFormProps) => {
+  const theme = useTheme();
+  const styles = getStyles(theme);
+
+  const handleSignInFormSubmit = async (credentials: Credentials) => {
+    try {
+      await PlatformLoginService.signIn(credentials);
+
+      getSettings();
+      showSuccessNotification({ message: `${Messages.loginSucceeded} ${credentials.email}` });
```
nicolalamacchia, line 23: I rename `loginSucceeded` as `signInSucceeded` to be consistend with sign up

pmm-app/src/pmm-settings/components/PlatformLogin/SignIn.tsx
```diff
@@ -0,0 +1,79 @@
...
+import { ButtonWithSpinner } from 'shared/components/Form';
+import { showErrorNotification, showSuccessNotification } from 'shared/components/helpers';
+import { Credentials, LoginFormProps } from './types';
+import { Messages } from './PlatformLogin.messages';
+import { InputFieldAdapter } from './FieldAdapters/FieldAdapters';
+import { getStyles } from './PlatformLogin.styles';
+import { PlatformLoginService } from './PlatformLogin.service';
+
+
+export const SignIn = ({ changeMode, getSettings }: LoginFormProps) => {
```
nicolalamacchia, line 14:
```suggestion
export const SignIn: FC<LoginFormProps> = ({ changeMode, getSettings }) => {
```

pmm-app/src/pmm-settings/components/PlatformLogin/SignUp.tsx
```diff
@@ -0,0 +1,111 @@
...
+import { ButtonWithSpinner } from 'shared/components/Form';
+import { showErrorNotification, showSuccessNotification } from 'shared/components/helpers';
+import { Credentials, LoginFormProps } from './types';
+import { Messages } from './PlatformLogin.messages';
+import { CheckboxFieldAdapter, InputFieldAdapter } from './FieldAdapters/FieldAdapters';
+import { getStyles } from './PlatformLogin.styles';
+import { PlatformLoginService } from './PlatformLogin.service';
+import { PRIVACY_POLICY_URL, TERMS_OF_SERVICE_URL } from './PlatformLogin.constants';
+
+export const SignUp = ({ changeMode, getSettings }: LoginFormProps) => {
```
nicolalamacchia, line 14:
```suggestion
export const SignUp: FC<LoginFormProps> = ({ changeMode, getSettings }) => {
```

nicolalamacchia (changes requested):

pmm-app/src/pmm-settings/components/PlatformLogin/SignUp.tsx
atymchuk: Yes, one field only for the password.

pmm-app/src/pmm-settings/components/PlatformLogin/SignUp.tsx
```diff
@@ -0,0 +1,111 @@
...
+      </ButtonWithSpinner>
+      <ButtonWithSpinner
+        data-qa="sign-up-to-sign-in-button"
+        className={styles.signInButton}
+        type="button"
+        variant="link"
+        onClick={changeMode}
+        disabled={submitting}
+      >
+        Back to login
```
nicolalamacchia, line 105: I think this should go in `Messages`

pmm-app/src/pmm-settings/components/PlatformLogin/SignUp.tsx
lunaticusgreen: done

nicolalamacchia (changes requested):

pmm-app/src/pmm-settings/components/PlatformLogin/PlatformLogin.test.tsx
```diff
@@ -1,29 +1,29 @@
 import React from 'react';
 import { mount } from 'enzyme';
-import { SignUp } from './SignUp';
+import { PlatformLogin } from './PlatformLogin';
 import { LoggedIn } from './LoggedIn/LoggedIn';
 
 jest.mock('shared/components/helpers/notification-manager');
 
 describe('SignUp::', () => {
```
tiagomotasantos, line 8: PlatformLogin::

pmm-app/src/pmm-settings/components/PlatformLogin/PlatformLogin.messages.ts
```diff
@@ -1,15 +1,18 @@
 export const Messages = {
-  signUp: 'Sign up',
+  perconaPlatform: 'Sign up',
```
tiagomotasantos, line 2: Sign Up

pmm-app/src/pmm-settings/components/PlatformLogin/PlatformLogin.messages.ts
```diff
@@ -1,15 +1,18 @@
...
   signIn: 'Sign in',
```
tiagomotasantos, line 3: Sign In (keeps it consistent with the other messages)

tiagomotasantos (changes requested):

pmm-app/src/pmm-settings/components/PlatformLogin/PlatformLogin.service.ts
```diff
@@ -0,0 +1,11 @@
+import { apiRequest } from 'shared/components/helpers/api';
+import { Credentials } from './types';
+
+export const PlatformLoginService = {
+  signUp(credentials: Credentials): Promise<void> {
+    return apiRequest.post<any, any>('/v1/Platform/SignUp', credentials);
+  },
+  signIn(credentials: Credentials): Promise<void> {
+    return apiRequest.post<any, any>('/v1/Platform/SignIn', credentials);
```
nicolalamacchia, line 9: There should be types here, I think `<void, Credentials>` should work :)

nicolalamacchia (changes requested):

pmm-app/src/pmm-settings/components/PlatformLogin/PlatformLogin.styles.ts
```diff
@@ -44,6 +44,7 @@ export const getStyles = stylesFactory((theme: GrafanaTheme) => ({
   `,
   form: css`
     max-width: 300px;
+    min-width: 300px;
```
nicolalamacchia, line 47: I would do something like `min-width: 150px` instead here and use `flex: 1` to `tabContentWrapper` in `Settings.styles.ts`, in order make this shrink a bit on smaller screens

nicolalamacchia (changes requested):

nicolalamacchia (approved):

PR #920, opened 2020-08-20
PMM-6500 db check style fix

atymchuk:


lunaticusgreen (approved):

tiagomotasantos (approved):

pmm-app/src/pmm-check/CheckPanel.tsx
```diff
@@ -113,14 +113,14 @@ export class CheckPanel extends PureComponent<CheckPanelProps, CheckPanelState>
...
-                {title ?? ''}
+                {title}
               </div>
               <ButtonWithSpinner
                 onClick={this.handleRunChecksClick}
                 isLoading={isRunChecksRequestPending}
                 disabled={hasNoAccess}
                 className={styles.runChecksButton}
-                variant="secondary"
+                variant="primary"
```
nicolalamacchia, line 123: I think the variant `primary` is the default, so this line could be removed

atymchuk: So, `explicit vs implicit` doesn't apply here? A go developer would argue, but I won't )

PR #937, opened 2020-08-27
PMM-6568 new modal

lunaticusgreen:
[PMM-6568](https://jira.percona.com/browse/PMM-6568)


pmm-app/src/shared/components/Elements/Modal/Modal.styles.ts
```diff
@@ -0,0 +1,54 @@
...
+      width: 100%;
+      max-height: calc(90vh - ${theme.spacing.d} * 2);
+  `,
+  modalHeaderClose: css`
+      height: 100%;
+      display: flex;
+      align-items: center;
+      flex-grow: 1;
+      justify-content: flex-end;
+      padding-right: ${theme.spacing.sm};
```
atymchuk, line 52: I guess it should be a margin - padding impacts the click area, which would be disproportionate in this case.

pmm-app/tests/QAN/pages/qanPage.js
```diff
@@ -64,7 +64,7 @@ module.exports = {
     showSelectedDisabled: '//button[@data-qa="qan-filters-show-selected" and @disabled ]',
     table: '//table//tr[2]',
     tableRow: 'div.tr',
-    noQueries: "//h1[contains(text(), 'No queries available for this combination of filters')]",
+    noQueries: '//h1[contains(text(), \'No queries available for this combination of filters\')]',
```
atymchuk, line 67: Let's make sure we avoid the escaped quotes `\'` - replace them with double quotes `"`. It should work.
```suggestion
    noQueries: '//h1[contains(text(), "No queries available for this combination of filters")]',
```

pmm-app/tests/pages/dashboardPage.js
```diff
@@ -543,8 +543,8 @@ module.exports = {
     clearSelection: '//a[@ng-click="vm.clearSelections()"]',
     Last2Days: '//span[contains(text(), \'Last 2 days\')]',
     timeRangePickerButton: '.btn.navbar-button.navbar-button--tight',
-    rootUser: "//div[contains(text(), 'root')]",
-    dataLinkForRoot: "//div[contains(text(), 'Data links')]/..//a",
+    rootUser: '//div[contains(text(), \'root\')]',
```
atymchuk, line 546: Ditto for the quotes.

pmm-app/src/shared/components/Elements/Modal/Modal.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+export const Modal: FC<ModalWindow> = (props) => {
+  const {
+    isVisible, children, title, onClose, closeOnClickaway = true, closeOnEscape = true
+  } = props;
+  const theme = useTheme();
+  const styles = getStyles(theme);
+
+  if (closeOnEscape) {
+    useEffect(() => {
```
atymchuk, line 23: How about this rule from https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level?
`Don’t call Hooks inside loops, conditions, or nested functions`. 

I wonder, doesn't eslint complain about that?

atymchuk (changes requested):

pmm-app/src/shared/components/Elements/Modal/Modal.tsx
lunaticusgreen: Good point, fixed

atymchuk (approved):

PR #938, opened 2020-08-28
PMM-6572 fix overlay height

lunaticusgreen:
[PMM-6572](https://jira.percona.com/browse/PMM-6572)

https://github.com/Percona-Lab/pmm-submodules/pull/1073


tiagomotasantos (changes requested):
Instead of adding `min-height` to the Overlay what about just removing `!loading` from Examples?
Because that was the point of the Overlay, always be on top of some content and keep the old content while loading.

lunaticusgreen:
> Instead of adding `min-height` to the Overlay what about just removing `!loading` from Examples?
> Because that was the point of the Overlay, always be on top of some content and keep the old content while loading.

@tiagomotasantos Main point of these changes was to add more flexibility to this component because sometimes we just may have situations of no content inside the overlay and at least we need to render spinner right

tiagomotasantos (approved):

PR #943, opened 2020-08-31
SAAS-295 filters fragment

yurkovychv:
[SAAS-295](https://jira.percona.com/browse/SAAS-295)


lunaticusgreen (approved):
👍

puneet0191 (approved):

puneet0191:
@yurkovychv please run nightly job on your branch

pmm-app/tests/QAN/steps/qanActions.js
```diff
@@ -176,7 +176,7 @@ module.exports = {
     assert.notEqual(
       countAfter,
       countBefore,
-      `After applying Filter value ${countBefore} should not be equal to ${countAfter}`,
+      `After applying a Filter, count of queries should not be equal to ${countAfter}`,
```
atymchuk, line 179: IMHO, we went from a more explicit to a less explicit error message. Now you have to research what is not equal to `countAfter`.

atymchuk (changes requested):

pmm-app/tests/QAN/steps/qanActions.js
yurkovychv: @atymchuk It was done on purpose actually. If the assertion fails it says: After applying Filter value 1 should not be equal to 1

atymchuk: lol

atymchuk (approved):

pmm-app/tests/QAN/steps/qanActions.js
puneet0191: lol good one, we don't really need to print values, I realized it is already available as part of expected and actual from Assert, we could just make it "After applying a Filter, count of queries was expected to change"

atymchuk: then double LOL

yurkovychv: yeah, we have a lot of such places

atymchuk: I'm learning something new every day. Will I ever stop?

puneet0191: if you stop learning, I think it would not be a good place to be in :stuck_out_tongue:

PR #966, opened 2020-09-10
PMM-6618 Modify clean_up script refer annotations

adivinho:
[PMM-6618](https://jira.percona.com/browse/PMM-6618)

### DIFF
--- a/misc/cleanup-dash.py
+++ b/misc/cleanup-dash.py
@@ -399,6 +399,11 @@ def add_annotation(dashboard):
     user_input = raw_input(prompt)
     if user_input:
         if user_input == 'Yes':
+            active_variables = [tag]
+            for templating in copy.deepcopy(dashboard['templating']['list']):
+                if templating['type'] == "query" and templating['hide'] == 0:
+                    active_variables.append('$' + templating['name'])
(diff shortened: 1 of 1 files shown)


misc/cleanup-dash.py
```diff
@@ -399,6 +399,11 @@ def add_annotation(dashboard):
     user_input = raw_input(prompt)
     if user_input:
         if user_input == 'Yes':
+            active_variables = [tag]
```
askomorokhov, line 402: Looks like we do not change `tag` here - can be a constant. Please use upper case for constants.

misc/cleanup-dash.py
```diff
@@ -399,6 +399,11 @@ def add_annotation(dashboard):
...
+            for templating in dashboard['templating']['list']:
+                if templating['type'] == "query" and templating['hide'] == 0:
```
askomorokhov, line 404:
```suggestion
                if templating['type'] == 'query' and templating['hide'] == 0:
```

Better to be consistent with `'` and `"`.

askomorokhov (approved):
Looks good to me with minor comments.

PR #980, opened 2020-09-28
PMM-6531 Change coloring of element Update Stats when Metadata Queried

adivinho:
[PMM-6531](https://jira.percona.com/browse/PMM-6531)

- [ ] https://github.com/Percona-Lab/pmm-submodules/pull/1133


dashboards/MySQL_InnoDB_Details.json
```diff
@@ -149,6 +161,11 @@
             "id": 59,
             "links": [],
             "mode": "html",
+            "options": {
+                "content": "<h1><i><font color=#5991A7><b><center>Data for </font><font color=#e68a00>$service_name</font>   <font color=#5991A7> running on </font>  </font><font color=#e68a00>$node_name</font> <font color=#5991A7> with</font>  </font><font color=#e68a00>$interval</font> <font color=#5991A7>resolution</center></b></font></i></h1>",
```
atymchuk, line 165: The `font` tag is [deprecated](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/font) since HTML 4.01, its use is discouraged by the HTML spec. 

Also, the code above suffers from a proper tag closing problem.

Consider applying a `style` css property on `span` tag: `<span style="color: #5991A7">....</span>`

atymchuk: Something like this would be more appropriate:
`<h1><i><b><center><span style="color:#5991A7">Data for </style><span style="color:#e68a00">$service_name</span><span style="color:#5991A7"> running on </span><span style="color:#e68a00">$node_name</span> <span style="color:#5991A7"> with</span> <span style="color:#e68a00">$interval</span> <span style="color:#5991A7">resolution</span></center></b></i></h1>`

atymchuk (changes requested):

dashboards/MySQL_InnoDB_Details.json
adivinho: I have removed this section from the PR. Also new ticket has been created because such element has to changed on many dashboards.

atymchuk (approved):

dashboards/MySQL_InnoDB_Details.json
atymchuk: Got it, approved.

PR #985, opened 2020-10-01
PMM-5975 List DB clusters

tiagomotasantos:
[PMM-5975](https://jira.percona.com/browse/PMM-5975)

This PR was built on top of PMM-5972 so the same FB can be used to test that one.

https://github.com/Percona-Lab/pmm-submodules/pull/1142


pmm-app/src/pmm-dbaas/components/XtraDB/XtraDB.hooks.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    try {
+      const clusters: XtraDBCluster[] = [];
+      const requests = kubernetes.map(XtraDBService.getXtraDBClusters);
+      const results = await processPromiseResults(requests);
+
+      results
+        .filter(filterFulfilled)
+        .map((r) => ((r as FulfilledPromiseResult).value.clusters)
+          .map(
```
lunaticusgreen, line 27: I'd say that using a push inside the map is not the best option, something like this looks simpler

` const clusters = results.filter(filterFulfilled).map(({value}) => toModel(value.clusters, DATABASE_LABELS.mysql)) `

tiagomotasantos: That wouldn't work, because "value.clusters" is a list of clusters and not a single cluster. We have to do one request per kubernetes cluster to get the DB clusters for that kubernetes cluster. That gives us a list of lists. A list of DB clusters per kubernetes cluster. I'm using the push there because in the end we want a single list and not a list of lists. Doing it this way prevents us from having to further iterate the list again to do that transformation. Tell me if you understand it @lunaticusgreen and if still think there's a better way pls.

lunaticusgreen: got it, thanks for the explanation 👍

lunaticusgreen (approved):

pmm-app/src/pmm-dbaas/components/XtraDB/__mocks__/XtraDB.hooks.ts
```diff
@@ -0,0 +1,18 @@
...
+export const useXtraDBClusters = (kubernetes: Kubernetes[]): [
+  XtraDBCluster[],
+  GetXtraDBClustersAction,
+  boolean
+] => {
+  const xtraDBClusters: XtraDBCluster[] = [];
+
+  if (kubernetes.length > 0) {
+    xtraDBClusters.push(...xtraDBClustersStub);
+  }
```
nicolalamacchia, line 15: this is faster:

```suggestion
  let xtraDBClusters: XtraDBCluster[] = [];

  if (kubernetes.length > 0) {
    xtraDBClusters = [...xtraDBClustersStub];
  }
```

pmm-app/src/pmm-dbaas/components/XtraDB/XtraDB.hooks.ts
```diff
@@ -0,0 +1,44 @@
...
+import { XtraDBCluster, GetXtraDBClustersAction } from './XtraDB.types';
+import { XtraDBService, toModel } from './XtraDB.service';
+
+export const useXtraDBClusters = (kubernetes: Kubernetes[]): [
+  XtraDBCluster[],
+  GetXtraDBClustersAction,
+  boolean
+] => {
+  const [xtraDBClusters, setXtraDBClusters] = useState<XtraDBCluster[]>([]);
+  const [loading, setLoading] = useState(false);
```
nicolalamacchia, line 14: I suggest to use something like `pendingRequest` instead of `loading`, in accordance to what we're doing for new code

nicolalamacchia (changes requested):
More tests are needed, especially for `XtraDB.hooks.ts`

pmm-app/src/pmm-dbaas/components/XtraDB/XtraDB.service.ts
```diff
@@ -1,28 +1,38 @@
 import { apiRequestManagement } from 'shared/components/helpers/api';
+import { Kubernetes } from '../Kubernetes/Kubernetes.types';
 import { XtraDBCluster, XtraDBClusterAPI } from './XtraDB.types';
 
 export const XtraDBService = {
+  getXtraDBClusters(kubernetes: Kubernetes) {
+    return apiRequestManagement.post<any, any>('/DBaaS/XtraDBClusters/List', kubernetes);
```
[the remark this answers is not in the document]
tiagomotasantos: Will take care of that in other PR.

PR #989, opened 2020-10-07
fix log offset update

lunaticusgreen:


pmm-app/src/pmm-update/hooks/usePerformUpdate.ts
```diff
@@ -49,7 +49,8 @@ export const usePerformUpdate = (): UpdateStatus => {
 
           return `${previousOutput}${logLines.join('\n')}\n`;
         });
-        newLogOffset = logOffset + log_offset ?? 0;
+        newLogOffset = log_offset ?? 0;
```
atymchuk, line 52: I see two major problems here:
1/ logOffset on line 39 overrides & mutates the same var declared on line 14. So we need to refactor it, otherwise it’s hardly even readable
2/ add a test to updateStatus

atymchuk (changes requested):

PR #996, opened 2020-10-12
PMM-6755 Enforce trailing comma

tiagomotasantos:
[PMM-6755](https://jira.percona.com/browse/PMM-6755)

Requires trailing comma when the last element is in a different line. 
Replaces `only-multiline` rule with `always-multiline` which enforces trailing comma.

### DIFF
--- a/pmm-app/.eslintrc
+++ b/pmm-app/.eslintrc
@@ -20,11 +20,11 @@
     "comma-dangle": [
       "error",
       {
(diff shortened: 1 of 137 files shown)


lunaticusgreen (approved):
👍

nicolalamacchia (approved):

atymchuk (approved):

PR #1002, opened 2020-10-15
PMM-6707 change the locator fetch logic for upgrade

puneet0191:
[PMM-6707](https://jira.percona.com/browse/PMM-6707)

This is to avoid using 2.9 for comparison rather use Jenkins parameter to fetch Major version being tested and load locator from that.

To avoid failure for DBaaS Cluster tests, since we can't use test  data anymore.


lunaticusgreen (approved):
👍

IvanaBrecherova (approved):

PR #1007, opened 2020-10-20
Changes for PMM 2.12

BupycHuk:


tiagomotasantos (approved):

tiagomotasantos:
Please rebase and merge (without squash), so we don't lose the git history.

PR #1025, opened 2020-10-30
PMM-6701 replace font tag by a style css property in dashboards

adivinho:
[PMM-6701](https://jira.percona.com/browse/PMM-6701)

- [ ] https://github.com/Percona-Lab/pmm-submodules/pull/1213


dashboards/Memory_Details.json
```diff
@@ -142,7 +142,7 @@
             "type": "digiapulssi-breadcrumb-panel"
         },
         {
-            "content": "<h1><i><font color=#5991A7><b><center>Data for </font><font color=#e68a00>$node_name</font> <font color=#5991A7> with</font> </font><font color=#e68a00>$interval</font> <font color=#5991A7>resolution</center></b></font></i></h1>",
+            "content": "<h1><i><b><center><span style=\"color:#5991A7\">Data for </span><span style=\"color:#e68a00\">$node_name</span> <span style=\"color:#5991A7\"> with</span> <span style=\"color:#e68a00\">$interval</span> <span style=\"color:#5991A7\">resolution</span></center></b></i></h1>",
```
atymchuk, line 145: Several major concerns here:
1/ the `center` tag is obsolete -> https://developer.mozilla.org/en-US/docs/Web/HTML/Element/center. Centering the text should be done with via the `text-align` style property;
2/ if we want the whole `h1` to have the color `#5991A7`, then there is no need to put so many spans; it's enough to move the style to the `h1` tag and remove the spans;
3/ there is no guarantee that `i`, `b` tags would work as expected; In fact, they do now, but may be overridden by a stylesheet; IMHO those should be styled via respective `font-style` and `font-weight` css properties;
4/ escaping double quotes with `\"` looks weird, there should be a better way to do it, i.e. try using single quotes for the outer string

The same comment would be valid for the rest of the PR, so I suggest you consult with @tiagomotasantos or @lunaticusgreen

atymchuk (changes requested):

dashboards/Memory_Details.json
adivinho: Blocks have been modified with changes refer to mentioned concerns.

atymchuk (approved):

PR #1031, opened 2020-11-03
PMM-6868 fix delete cluster confirmation text

lunaticusgreen:
[PMM-6868](https://jira.percona.com/browse/PMM-6868)


pmm-app/src/pmm-dbaas/DBaaS.messages.ts
```diff
@@ -11,7 +11,7 @@ export const Messages = {
     deleteModal: {
```
tiagomotasantos, line 11: Create `deleteModal` section inside `dbcluster` section and use it in `DeleteDBClusterModal.tsx`, because this is being used in two different places and it shouldn't.

tiagomotasantos (changes requested):

pmm-app/src/pmm-dbaas/DBaaS.messages.ts
lunaticusgreen: Thanks, created separate fields

tiagomotasantos (approved):

PR #1037, opened 2020-11-05
PMM-6856 Add icon to go to dashboard

tiagomotasantos:
[PMM-6856](https://jira.percona.com/browse/PMM-6856)


nicolalamacchia (approved):

pmm-app/src/pmm-dbaas/components/DBCluster/DBClusterName/DBClusterName.tsx
```diff
@@ -0,0 +1,31 @@
...
+    `${DASHBOARD_URL_MAP[databaseType]}${clusterName}`
+  ), [databaseType, clusterName]);
+
+  return (
+    <div className={styles.clusterNameWrapper}>
+      <span>{clusterName}</span>
+      <a
+        href={dashboardURL}
+        target="_blank"
+        rel="noreferrer"
```
atymchuk, line 24:
```suggestion
        rel="noreferrer noopener"
```
That's what MDN suggest for more security

atymchuk (changes requested):

pmm-app/src/pmm-dbaas/components/DBCluster/DBClusterName/DBClusterName.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { DASHBOARD_URL_MAP } from './DBClusterName.constants';
+
+export const DBClusterName: FC<DBClusterNameProps> = ({
+  dbCluster: {
+    clusterName,
+    databaseType,
+  },
+}) => {
+  const styles = useStyles(getStyles);
+  const dashboardURL = useMemo(() => (
```
atymchuk, line 14: nit: while there is no harm, there is probably no benefit either, since this computation is not expensive. 
Quote: "This optimization helps to avoid expensive calculations on every render."
Reference: https://reactjs.org/docs/hooks-reference.html#usememo
One of other possible use cases of it is when we pass it to a child as a prop. Then it'll prevent the child from re-rendering.

atymchuk (approved):

pmm-app/src/pmm-dbaas/components/DBCluster/DBClusterName/DBClusterName.tsx
tiagomotasantos: True, thanks! I'm working on PMM-6843 now so will remove it there.

PR #1054, opened 2020-11-26
PMM-6986 allow deletion for cluster that is in pending state

lunaticusgreen:
[PMM-6986](https://jira.percona.com/browse/PMM-6986)


pmm-app/src/pmm-dbaas/components/DBCluster/DBClusterActions/DBClusterActions.test.tsx
```diff
@@ -37,25 +37,43 @@ describe('DBClusterActions::', () => {
 
     expect(root.find('button').prop('disabled')).toBeFalsy();
   });
-  it('disables button if cluster changing', () => {
+
+  it('calls delete action correctly', async () => {
```
tiagomotasantos, line 41: IMO we can still have a test to check if other actions are disabled when the cluster is changing. For example the restart.

tiagomotasantos (approved):

PR #1061, opened 2020-12-01
remove public url part

lunaticusgreen:


pmm-app/src/pmm-dbaas/components/DBCluster/DBCluster.tsx
```diff
@@ -80,9 +80,9 @@ export const DBCluster: FC<DBClusterProps> = ({ kubernetes }) => {
     () => (
       <AddClusterButton
         label={Messages.dbcluster.addAction}
-        disabled={settingsLoading || !settings?.publicAddress}
-        showWarning={!settingsLoading && !settings?.publicAddress}
-        warningMessage={buildWarningMessage(styles.settingsLink)}
+        // disabled={settingsLoading || !settings?.publicAddress}
```
BupycHuk, line 83: maybe it's better to delete only `publicAddress`?

lunaticusgreen: We don't need disabled anymore because we only need it for address check

BupycHuk (approved):

PR #1093, opened 2021-01-15
PMM-7048 Remove create btn from second step

tiagomotasantos:
[PMM-7048](https://jira.percona.com/browse/PMM-7048) [&#10088;?&#10089;](https://pullrequestbadge.com/?utm_medium=github&utm_source=percona&utm_campaign=badge_info)

Removes create cluster button from step two and puts it outside the steps so the user can create cluster without going to step two. For that it was necessary to fork the `StepProgress` component as discussed and specified in the ticket, which makes sense to adapt it for this specific DBaaS case (and more specific changes might be done here).
Also fixes required database type to be able to create cluster.

https://github.com/Percona-Lab/pmm-submodules/pull/1395


lunaticusgreen (approved):
👍

PR #1102, opened 2021-01-21
PMM-7320 Alert rules UI tests

yurkovychv:
[PMM-7320](https://jira.percona.com/browse/PMM-7320) [&#10088;?&#10089;](https://pullrequestbadge.com/?utm_medium=github&utm_source=percona&utm_campaign=badge_info)

### DIFF
--- a/pmm-app/tests/ia/alertRules_test.js
+++ b/pmm-app/tests/ia/alertRules_test.js
@@ -4,7 +4,7 @@ const page = require('./pages/alertRulesPage');
 const rules = new DataTable(['template', 'templateType', 'ruleName', 'threshold', 'duration',
   'severity', 'filters', 'channels', 'activate']);
 
-Object.values(page.rules).forEach(rule => {
+Object.values(page.rules).forEach((rule) => {
   rules.add([rule.template, rule.templateType, rule.ruleName, rule.threshold, rule.duration,
     rule.severity, rule.filters, rule.channels, rule.activate]);
 });
--- a/pmm-app/tests/ia/alertRules_test.js
+++ b/pmm-app/tests/ia/alertRules_test.js
@@ -1,0 +1,160 @@
+const assert = require('assert');
+const page = require('./pages/alertRulesPage');
+
+const rules = new DataTable(['template', 'templateType', 'ruleName', 'threshold', 'duration',
+  'severity', 'filters', 'channels', 'activate']);
+
+for (const i in page.rules) {
+  const rule = page.rules[i];
+  rules.add([rule.template, rule.templateType, rule.ruleName, rule.threshold, rule.duration,
+    rule.severity, rule.filters, rule.channels, rule.activate])
+}
+
+Feature('IA: Alert rules');
+
+Before(async (I, alertRulesPage, settingsAPI) => {
+  I.Authorize();
+  await settingsAPI.apiEnableIA();
+});
+
+BeforeSuite(async (I, alertRulesPage, settingsAPI, rulesAPI, templatesAPI, channelsAPI, ncPage) => {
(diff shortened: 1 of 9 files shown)


puneet0191 (approved):

pmm-app/tests/ia/alertRules_test.js
```diff
@@ -0,0 +1,162 @@
+const assert = require('assert');
+const page = require('./pages/alertRulesPage');
+
+const rules = new DataTable(['template', 'templateType', 'ruleName', 'threshold', 'duration',
+  'severity', 'filters', 'channels', 'activate']);
+
+for (const i in page.rules) {
```
atymchuk, line 7: nit: could be better if we use a more modern API like `Object.values(page.rules).forEach(rule => ...)`. This way you don't need to bother with indices :)

yurkovychv: sure, fixing it. Thanks

pmm-app/tests/ia/pages/api/channelsAPI.js
```diff
@@ -60,6 +64,7 @@ module.exports = {
     const body = { channel_id: channelId };
     const resp = await I.sendPostRequest('v1/management/ia/Channels/Remove', body, headers);
 
-    assert.ok(resp.status === 200, `Failed to remove channel with channel_id=${channelId}`);
+    assert.ok(resp.status === 200, `Failed to remove channel with channel_id "${channelId}".
+     Response message is "${resp.data.message}"`);
```
atymchuk, line 68:
```suggestion
    assert.ok(
      resp.status === 200, 
      `Failed to remove channel with channel_id "${channelId}". Response message is "${resp.data.message}"`
    );
```
This is how to approach similar long lines :)

pmm-app/tests/ia/pages/api/templatesAPI.js
```diff
@@ -0,0 +1,45 @@
...
+     Response message is "${resp.data.message}"`);
+
+    return id;
+  },
+
+  async clearAllTemplates() {
+    const headers = { Authorization: `Basic ${await I.getAuth()}` };
+    const resp = await I.sendPostRequest('v1/management/ia/Templates/List', { reload: true }, headers);
+
+    if (resp.data === {}) return;
```
atymchuk, line 23: This will never be true, since the object `{}` gets re-created on every pass. In this case we should check like this `Object.keys(resp.data).length === 0` - this is reliable.

pmm-app/tests/ia/pages/api/templatesAPI.js
```diff
@@ -0,0 +1,45 @@
...
+
+    for (const i in resp.data.templates) {
+      const template = resp.data.templates[i];
+
+      // eslint-disable-next-line no-continue
+      if (template.source === 'BUILT_IN') continue;
```
atymchuk, line 29: LOL, continue is old-school, you could as well do this:
`
if (template.source !== 'BUILT_IN') {
   await this.removeTemplate(template.name);
}
`

atymchuk: Then you don't need to continue :)

atymchuk (changes requested):

pmm-app/tests/ia/pages/api/templatesAPI.js
yurkovychv: done, thanks

pmm-app/tests/ia/pages/api/templatesAPI.js
yurkovychv: done, thanks

atymchuk (approved):

PR #1118, opened 2021-02-02
PMM-6513 Update UI tests for new PMM Settings page

IvanaBrecherova:


IvanaBrecherova:
https://pmm.cd.percona.com/blue/organizations/jenkins/pmm2-ui-tests/detail/pmm2-ui-tests/6706/tests - Tests are running correctly.

pmm-app/tests/pages/pmmSettingsPage.js
```diff
@@ -238,10 +247,18 @@ module.exports = {
 
       I.refreshPage();
       I.wait(5);
+
+      if (!added && i === 5) {
```
atymchuk, line 251: What is this magic number about? In such cases we should have a comment, since you can't infer it from the code.

atymchuk: One more: if we break at `i == 5`, why do we need to set ` i < 30`?

pmm-app/tests/verifyPMMSettingsPageElements_test.js
```diff
@@ -1,15 +1,16 @@
 const assert = require('assert');
 const page = require('./pages/pmmSettingsPage');
 
-const validationValues = ['text ', '2147483648', '-1', '<script>alert(test);</script>'];
+const validationValues = ['2147483648', '-1', '0'];
```
atymchuk, line 4: Not sure again what this number means `2147483648 `. Need comments please.

atymchuk (changes requested):

pmm-app/tests/verifyPMMSettingsPageElements_test.js
IvanaBrecherova: @atymchuk Just testing values: I followed this test case: https://jira.percona.com/secure/Tests.jspa#/testCase/97 :)

pmm-app/tests/pages/pmmSettingsPage.js
IvanaBrecherova: My explanation: I've added that condition because if I am checking that rule is NOT added I think that I don't need to wait 150 seconds for a response. So I am waiting just 30 seconds for NOT added rule. I think that 30 iterations were added for a really slow response from API. I will add a comment to the code but If you want a better explanation we can call each other. :)

yurkovychv:
@IvanaBrecherova please rebase to release/2.15

pmm-app/tests/verifyPMMSettingsPageFunctionality_test.js
```diff
@@ -39,21 +39,32 @@ Scenario('Open PMM Settings page and verify changing Data Retention [critical]',
...
+    await pmmSettingsPage.verifyPopUpMessage(pmmSettingsPage.messages.successPopUpMessage);
+    pmmSettingsPage.openAlertsManagerUi();
+    await pmmSettingsPage.verifyAlertmanagerRuleAdded(pmmSettingsPage.alertManager.editRuleName);
+    // PMM-T109 starting here
+    I.amOnPage(pmmSettingsPage.url);
+    await pmmSettingsPage.waitForPmmSettingsPageLoaded();
+    await pmmSettingsPage.expandSection(sectionNameToExpand, pmmSettingsPage.fields.alertmanagerButton);
+    pmmSettingsPage.addAlertmanagerRule('', '');
+    pmmSettingsPage.openAlertsManagerUi();
+    I.wait(10);
```
puneet0191, line 64: no other option?

puneet0191 (approved):

pmm-app/tests/pages/pmmSettingsPage.js
IvanaBrecherova: Outdated :)

pmm-app/tests/verifyPMMSettingsPageElements_test.js
```diff
@@ -1,14 +1,46 @@
 const assert = require('assert');
+const page = require('./pages/pmmSettingsPage');
+
+const validationValues = ['2147483648', '-1', '0'];
```
atymchuk, line 4:
```suggestion
// Value should be in range from 1 to 3650 days, so put a value outside of the range
const validationValues = ['2147483648', '-1', '0'];
```

atymchuk: That's what I meant by `need a comment`

atymchuk (changes requested):

atymchuk (approved):

puneet0191 (approved):

puneet0191 (changes requested):

puneet0191 (approved):

PR #1125, opened 2021-02-09
PMM-6704 Adding DB Cluster Page Validation & Tests

puneet0191:
Will only be merged after PMM-7420


pmm-app/tests/DbaaS/verifyDBaaS_test.js
```diff
@@ -0,0 +1,242 @@
...
+    I.amOnPage(dbaasPage.url);
+    dbaasPage.checkCluster(clusterName, false);
+    I.dontSeeElement(adminPage.fields.timePickerMenu);
+    I.click(dbaasPage.tabs.dbClusterTab.dbClusterTab);
+    I.waitForElement(dbaasPage.tabs.dbClusterTab.addDbClusterButton, 30);
+    I.dontSeeElement(adminPage.fields.timePickerMenu);
+    I.click(dbaasPage.tabs.dbClusterTab.addDbClusterButton);
+    I.waitForElement(dbaasPage.tabs.dbClusterTab.monitoringWarningLocator, 30);
+    I.waitForText(dbaasPage.monitoringWarningMessage, 30);
+    await adminPage.verifyBackgroundColor(dbaasPage.tabs.dbClusterTab.optionsCountLocator(1), 'rgb(235, 123, 24)');
```
yurkovychv, line 158: I don't think it is a good idea to check the exact colors here. It should be enough to check that colorAfter some action !== to colorBefore action if this applicable

pmm-app/tests/DbaaS/verifyDBaaS_test.js
```diff
@@ -0,0 +1,242 @@
...
+    assert.strictEqual(
+      await I.grabAttributeFrom(dbaasPage.tabs.dbClusterTab.advancedOptions.fields.cpuNumberFields, 'disabled'),
+      true,
+      'Number of CPU Field Should be disabled',
+    );
+    assert.strictEqual(
+      await I.grabAttributeFrom(dbaasPage.tabs.dbClusterTab.advancedOptions.fields.diskSizeInputField, 'disabled'),
+      true,
+      'Disk Size field must be disabled',
+    );
```
yurkovychv, line 208:
```suggestion
    assert.ok(
      await I.grabAttributeFrom(dbaasPage.tabs.dbClusterTab.advancedOptions.fields.diskSizeInputField, 'disabled'),
      'Disk Size field must be disabled',
    );
```
I think this way it should also work.

atymchuk (changes requested):

pmm-app/tests/DbaaS/api/dbaasAPI.js
```diff
@@ -0,0 +1,110 @@
...
+    const headers = { Authorization: `Basic ${await I.getAuth()}` };
+
+    const response = await I.sendPostRequest('v1/management/DBaaS/Kubernetes/List', body, headers);
+
+    if (typeof response.data.kubernetes_clusters !== 'undefined') {
+      const cluster = response.data.kubernetes_clusters.find(
+        (o) => o.kubernetes_cluster_name === clusterName,
+      );
+
+      return typeof cluster !== 'undefined';
```
atymchuk, line 54:
```suggestion
      return response.data.kubernetes_clusters.includes(clusterName);
```

atymchuk: That'd be much shorter )

pmm-app/tests/DbaaS/api/dbaasAPI.js
```diff
@@ -0,0 +1,110 @@
...
+    );
+  },
+
+  async apiCheckRegisteredClusterExist(clusterName) {
+    const body = JSON.stringify({});
+    const headers = { Authorization: `Basic ${await I.getAuth()}` };
+
+    const response = await I.sendPostRequest('v1/management/DBaaS/Kubernetes/List', body, headers);
+
+    if (typeof response.data.kubernetes_clusters !== 'undefined') {
```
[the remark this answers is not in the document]
atymchuk: This would work as well `if (response.data.kubernetes_clusters) {...}`

atymchuk (changes requested):

pmm-app/tests/pages/adminPage.js
```diff
@@ -111,4 +111,20 @@ module.exports = {
...
+    const backgroundColor = await I.grabCssPropertyFrom(element, 'background-color');
+
+    assert.strictEqual(
+      backgroundColor[0],
+      expectedColor,
+      `The Background color of the ${element} is not matching, the actual value is ${backgroundColor}`,
+    );
+  },
+
+  customClearField(field) {
```
yurkovychv, line 125: In some places just a simple `I.clearField()` works great, have you tried it?

pmm-app/tests/DbaaS/api/dbaasAPI.js
puneet0191: I am not sure, because it is an array of objects and I am first searching for the object with a matching cluster name.

pmm-app/tests/pages/adminPage.js
puneet0191: tried, doesn't work here, will check again.

pmm-app/tests/DbaaS/verifyDBaaS_test.js
puneet0191: hmm I can do that too, but what is wrong with checking the RGB value? it is still verification of color code which we expect to see, doesn't matter how it looks like on the UI/browser

pmm-app/tests/DbaaS/verifyDBaaS_test.js
puneet0191: sure, makes sense I will update

pmm-app/tests/DbaaS/verifyDBaaS_test.js
yurkovychv: IMHO it does not look like something we should cover with UI tests. I mean assert the exact color.

puneet0191: but it is not a visual verification, it is a verification of an expected RGB value, I don't know why it should not be covered, probably when it is changed, we only need to change the expected RGB value here..it is not complex to verify this, and ultimately this is part of user experience while trying to use the app -> UI. 

so look at it like we check for enabled/disabled fields (toggle button etc) in the UI, if we can do that, why we should skip this?

yurkovychv: just a suggestion. I had a similar test, which started failing because the color changed. So it was causing red job, and some distractions.

puneet0191: sure I understand your point

pmm-app/tests/DbaaS/api/dbaasAPI.js
```diff
@@ -0,0 +1,110 @@
...
+    for (let i = 0; i < 30; i++) {
+      const response = await I.sendPostRequest('v1/management/DBaaS/XtraDBClusters/List', body, headers);
+
+
+      if (typeof response.data.clusters !== 'undefined') {
+        const cluster = response.data.clusters.find(
+          (o) => o.name === dbClusterName,
+        );
+
+        if (typeof cluster === 'undefined') {
```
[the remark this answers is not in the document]
puneet0191: done

pmm-app/tests/DbaaS/api/dbaasAPI.js
```diff
@@ -0,0 +1,110 @@
...
+
+    for (let i = 0; i < 30; i++) {
+      const response = await I.sendPostRequest('v1/management/DBaaS/XtraDBClusters/List', body, headers);
+
+      if (typeof response.data.clusters !== 'undefined') {
+        const cluster = response.data.clusters.find(
+          (o) => o.name === dbClusterName,
+        );
+
+        if (typeof cluster !== 'undefined' && cluster.state === 'XTRA_DB_CLUSTER_STATE_READY') {
```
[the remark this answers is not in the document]
puneet0191: done

pmm-app/tests/DbaaS/pages/dbaasPage.js
```diff
@@ -0,0 +1,151 @@
...
+  monitoringWarningMessage: 'If you want to use monitoring, you need to set your PMM installation public address in',
+  requiredFieldError: 'Required field',
+  valueGreatThanErrorText: (value) => `Value should be greater or equal to ${value}`,
+  tabs: {
+    kubernetesClusterTab: {
+      addKubernetesClusterButton: '$kubernetes-new-cluster-button',
+      addKubernetesClusterButtonInTable: '//div[@data-qa=\'table-no-data\']//span[contains(text(), \'Register new Kubernetes Cluster\')]',
+      actionsLocator: (clusterName) => `//td[contains(text(), "${clusterName}")]//parent::tr//button[@data-qa="dropdown-menu-toggle"]`,
+      closeButton: '$modal-close-button',
+      clusterConfigurationText: '//div[@data-qa=\'pmm-overlay-wrapper\']//pre',
```
yurkovychv, line 19: also, could you please replace single quotes escaping ( `\'` ) to double quotes ( `"` )

puneet0191: Done

pmm-app/tests/DbaaS/verifyDBaaS_test.js
puneet0191: Done

pmm-app/tests/DbaaS/verifyDBaaS_test.js
puneet0191: Resolving, for now, will keep it like this, probably skip/remove the check-in case it fails a lot due to this change.

yurkovychv (dismissed):

pmm-app/tests/DbaaS/api/dbaasAPI.js
atymchuk: It will, it does exactly the same. Pls try, since we tend to have a lot of copy/paste, and that sets out a good example of how to solve it.

pmm-app/tests/DbaaS/api/dbaasAPI.js
```diff
@@ -0,0 +1,110 @@
...
+
+
+      if (response.data.clusters) {
+        const cluster = response.data.clusters.find(
+          (o) => o.name === dbClusterName,
+        );
+
+        if (cluster === undefined) {
+          break;
+        }
```
atymchuk, line 104:
```suggestion
        if (!response.data.clusters.includes(dbClusterName) {
            break;
        }
```

atymchuk: Same as above.

atymchuk (changes requested):

pmm-app/tests/DbaaS/api/dbaasAPI.js
puneet0191: I have tried it..it doesn't work and I am not sure if I am confused here 
```
const kubernetes_clusters = [
  {
    kubernetes_cluster_name: 'Cluster1',
    operators: { xtradb: [Object], psmdb: [Object] },
    status: 'KUBERNETES_CLUSTER_STATUS_OK'
  },
  {
    kubernetes_cluster_name: 'Testing',
    operators: { xtradb: [Object], psmdb: [Object] },
    status: 'KUBERNETES_CLUSTER_STATUS_OK'
  },
  {
    kubernetes_cluster_name: 'Kubernetes_Testing_Cluster_Minikube',
    operators: { xtradb: [Object], psmdb: [Object] },
    status: 'KUBERNETES_CLUSTER_STATUS_OK'
  }
];
```

That's the structure of kubernetes_clusters and if the array contains strings/numbers then I think array.includes would work just fine, but in this case, we need to filter by specific property of the object, hence I am using find and checking property `kubernetes_cluster_name`

atymchuk: Correct, that wouldn't work in your case ))

atymchuk (dismissed):

yurkovychv (approved):

PR #1160, opened 2021-02-22
PMM-7556 remove pmm-settings panel

lunaticusgreen:


pmm-app/src/plugin.json
```diff
@@ -23,14 +23,6 @@
   "includes": [
     { "type": "dashboard", "name": "pmm-qan", "path": "dashboards/pmm-qan.json" },
     { "type": "panel", "name": "pmm-qan-app-panel" },
-    { "type": "dashboard", "name": "PMM Inventory", "path": "dashboards/pmm-inventory.json" },
```
tiagomotasantos, line 26: Why remove this other panels too?

lunaticusgreen: That's because in this task we're moving 4 panels: 
1) Add instance
2) Settings
3) Inventory
4) Stt

PR #1178, opened 2021-03-15
PMM-7633

IvanaBrecherova:


pmm-app/tests/verifyPMMInventory_test.js
```diff
@@ -154,7 +154,8 @@ Scenario(
     const countOfRunning = await pmmInventoryPage.getCountOfRunningAgents();
     // Need countOfPMMAgentType because agents that have Type PMM Agent don't have status. We need subtract it
     const countOfPMMAgentType = await pmmInventoryPage.getCountOfPMMAgents();
+    const serviceIdsNotRunning = await pmmInventoryPage.getNotRunningAgentsServiceId();
```
puneet0191, line 157: So we are sure we will always have Agents in Waiting, Starting and Unknown status?

pmm-app/tests/pages/pmmInventoryPage.js
```diff
@@ -65,6 +65,15 @@ module.exports = {
     }
   },
 
+  async getNotRunningAgentsServiceId() {
+    let serviceIDs = await I.grabTextFromAll('//span[contains(text(), \'status: WAITING\')]/preceding-sibling::span[contains(text(), \'service_id:\')]');
```
puneet0191, line 69: I think you can improve the locator and since you use the same locator at 3 places, this is not a really an efficient way to handle the locator for Agent Status, try to make it more generic and please use `locate` it would make this readable

puneet0191 (changes requested):

pmm-app/tests/verifyPMMInventory_test.js
IvanaBrecherova: I've changed it to the array of statuses. We can easily delete or add new status in the future if it will be needed. :)

yurkovychv (approved):

yurkovychv (dismissed):

puneet0191 (approved):

yurkovychv (approved):

PR #1206, opened 2021-04-02
PMM-7739 - tests

yurkovychv:


puneet0191 (approved):

pmm-app/tests/ia/pages/alertRulesPage.js
```diff
@@ -1,5 +1,5 @@
 const { I } = inject();
-const rulesNameCell = (ruleName) => `//td[1][div[text()="${ruleName}"]]`;
+const rulesNameCell = (ruleName) => `//td[1][div/span[text()="${ruleName}"]]`;
```
puneet0191, line 2: can't use locate?

yurkovychv: no, I use this func in building another locators. For some reason, strange exceptions occur. So using this for now

PR #1212, opened 2021-04-08
Fix False positive

puneet0191:


yurkovychv:
https://github.com/percona/grafana-dashboards/pull/1211

PR #1213, opened 2021-04-08
PMM-7671 - Create UI tests for Remote Postgresql Instances from RDS

IvanaBrecherova:


IvanaBrecherova:
Moved to new repository

PR #1216, opened 2021-04-12
PMM-7586 Create tests for checking add HAProxy via UI

IvanaBrecherova:


pmm-app/tests/verifyRemoteInstances_test.js
```diff
@@ -113,3 +113,66 @@ Scenario(
...
+    I, remoteInstancesPage, pmmInventoryPage,
+  }) => {
+    const serviceName = 'haproxy_remote';
+    const url = process.env.PMM_UI_URL;
+
+    I.amOnPage(remoteInstancesPage.url);
+    remoteInstancesPage.waitUntilRemoteInstancesPageLoaded();
+    remoteInstancesPage.openAddRemotePage('haproxy');
+    I.waitForVisible(remoteInstancesPage.fields.hostName, 30);
+    I.fillField(remoteInstancesPage.fields.hostName, url.slice(7, url.length - 1));
```
yurkovychv, line 163: Maybe it's better to use [`URL`](https://developer.mozilla.org/ru/docs/Web/API/URL/URL) here, and use `url.host` instead ?

IvanaBrecherova: I don't understand it. If you mean `url.slice(7, url.length - 1)` change to `url.host` so that's doesn't work.

yurkovychv: It doesn't work because in your case `url` is not of type `URL`.
First you should create it like this: 
```suggestion
    const url = new URL(stringUrl);
```

and then use `url.host`

IvanaBrecherova: Thanks! :) Great, that's working! :)

IvanaBrecherova:
Moved to the new repository.

PR #1253, opened 2021-08-10
PMM-8509 Disable check updates

tiagomotasantos:
Shows messages when check for updates is disabled

https://github.com/percona-platform/grafana/pull/207
https://github.com/Percona-Lab/pmm-submodules/pull/1939

### DIFF
--- a/pmm-app/.eslintrc
+++ b/pmm-app/.eslintrc
@@ -89,5 +89,6 @@
     "no-unused-expressions": ["off"],
     "class-methods-use-this": ["off"],
     "react/require-default-props": ["off"],
+    "no-nested-ternary": ["off"],
   }
 }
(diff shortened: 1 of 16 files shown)


pmm-app/.eslintrc
```diff
@@ -89,5 +89,6 @@
     "no-unused-expressions": ["off"],
     "class-methods-use-this": ["off"],
     "react/require-default-props": ["off"],
+    "no-nested-ternary": ["off"],
```
tiagomotasantos, line 92: Disabled this because I think there's no better way of doing it, see `InfoBox.tsx`

fabio-silva (approved):

fabio-silva (dismissed):

fabio-silva (approved):

pmm-app/.eslintrc
atymchuk: > Disabled this because I think there's no better way of doing it, see `InfoBox.tsx`

Actually there is always a way to do it, please check out 
Multi-storey ternary is just too diffcult to read and comprehend )

PR #1260, opened 2021-08-30
PMM-8509 revert the lint rule back to more restrictive

atymchuk:
Fixes in this PR:
1/ put the `no-nested-ternary` back on
2/ updated the tests with `mount` vs `shallow`
3/ updated the contributing guide
4/ deleted redundant root files - package.lock and docker-compose.yml


fabio-silva (approved):

pmm-app/src/pmm-update/components/InfoBox/InfoBox.tsx
```diff
@@ -13,27 +23,47 @@ export const InfoBox: FC<InfoBoxProps> = ({
 }) => {
   const styles = useStyles(getStyles);
 
-  return (
-    <section data-qa="updates-info" className={styles.infoBox}>
-      {hasNoAccess ? (
+  if (hasNoAccess) {
```
tiagomotasantos, line 26: I don't see a clear advantage of having it this way besides the fact that we don't need to turn off the rule. This way it needs more code and multiple returns (which might be a bad practice as well). Anyway either one is fine, this is just a minor detail, so let's go!

tiagomotasantos (approved):

pmm-app/src/pmm-update/components/InfoBox/InfoBox.tsx
atymchuk: Arguably a bad practice, the code is much easier to read and therefore to maintain.

atymchuk: We do not want to save on the lines of code, we do want to be able to understand how it works. And the nested ternary fails short on that end.

atymchuk: If a particular eslint rule is bad or does more harm than good, I suggest we convince the community. Then it will be objective.

PR #1271, opened 2021-09-29
PMM-8541: fix link from Home dashboard to settings via CheckPanel

fabio-silva:


tiagomotasantos (approved):

pmm-app/src/pmm-check-home/components/Failed/Failed.tsx
```diff
@@ -30,9 +29,7 @@ export const Failed: FC<FailedProps> = ({ failed = [0, 0, 0], isSttEnabled, hasN
         Security Threat Tool is disabled.
         <br />
         {'Check '}
-        <Link className={styles.Link} to={PMM_SETTINGS_URL}>
-          PMM Settings.
-        </Link>
```
solovevayaroslavna, line 35: Hi! I'm new to the project. Could you explain why you decided to remove react-router-dom and history? (This is not a remark, I just want to understand the purpose of the changes)

fabio-silva: Hey. Well, these dashboards live within Grafana, which means they are already wrapped in a router (in Grafana's case, it's still Angular's, should be React's soon). I believe having this Router here was the cause of the bug: after clicking the link, it would take around 30s for the navigation to happen.

solovevayaroslavna: Thanks for the explanation!

solovevayaroslavna (approved):

PR #1278, opened 2021-10-22
Update convert-dash-from-PMM.py to clarify it's usage 

rnovikovP:
Note: https://github.com/percona/grafana-dashboards/blob/main/misc/convert-dash-to-PMM.py also need to have a similar header

### DIFF
--- a/misc/convert-dash-from-PMM.py
+++ b/misc/convert-dash-from-PMM.py
@@ -1,3 +1,19 @@
+##############################################################################################################
+# This script converts a PMM dashboard so it can be used in an external Prometheus + Grafana installation. 
+# It doesn’t need any input from you. It replaces PMM2 labels (node_name, service_name) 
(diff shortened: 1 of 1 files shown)


sybbjs:
some error here when format MongoDB_Instances_Overview.json:

Traceback (most recent call last):
  File "convert-dash-from-PMM.py", line 160, in <module>
    main()
  File "convert-dash-from-PMM.py", line 143, in main
    check_formulas(dashboard, "node_name", "instance")
  File "convert-dash-from-PMM.py", line 92, in check_formulas
    expr = re.sub('node_type=~\"[a-z,$_|]*\"', '', expr)
  File "/usr/lib64/python2.7/re.py", line 151, in sub
    return _compile(pattern, flags).sub(repl, string, count)
TypeError: expected string or buffer

PR #1297, opened 2021-12-31
PMM-7031 Fix API helper types

tiagomotasantos:
Fixes API helper types by removing the void return from functions and leaving just the generic type.

https://jira.percona.com/browse/PMM-7031

### DIFF
--- a/pmm-app/src/shared/components/helpers/api.ts
+++ b/pmm-app/src/shared/components/helpers/api.ts
@@ -10,17 +10,13 @@ export class ApiRequest {
     });
   }
 
-  async get<T, B>(path: string, query?: { params: B; cancelToken?: CancelToken }): Promise<void | T> {
+  async get<T, B>(path: string, query?: { params: B; cancelToken?: CancelToken }): Promise<T> {
     return this.axiosInstance
       .get<T>(path, query)
       .then((response): T => response.data)
-      .catch((e): void => {
+      .catch((e) => {
         showErrorNotification({ message: e.message });
-        if (axios.isCancel(e)) {
-          throw e;
-        } else {
-          showErrorNotification({ message: e.message });
-        }
+        throw e;
       });
   }
 
@@ -42,21 +38,21 @@ export class ApiRequest {
       });
   }
 
-  async delete<T>(path: string): Promise<void | T> {
+  async delete<T>(path: string): Promise<T> {
     return this.axiosInstance
       .delete<T>(path)
       .then((response): T => response.data)
-      .catch((): void => {
-        // Notify.error(e.message);
+      .catch((e) => {
+        throw e;
       });
   }
 
-  async patch<T, B>(path: string, body: B): Promise<void | T> {
+  async patch<T, B>(path: string, body: B): Promise<T> {
     return this.axiosInstance
       .patch<T>(path, body)
       .then((response): T => response.data)
-      .catch((): void => {
-        // Notify.error(e.message);
+      .catch((e) => {
+        throw e;
       });
   }
 }
--- a/pmm-app/src/shared/components/helpers/api.ts
+++ b/pmm-app/src/shared/components/helpers/api.ts
@@ -43,6 +43,7 @@ export class ApiRequest {
       .delete<T>(path)
       .then((response): T => response.data)
       .catch((e) => {
+        showErrorNotification({ message: e.message });
         throw e;
       });
   }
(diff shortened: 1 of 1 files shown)


pmm-app/src/shared/components/helpers/api.ts
```diff
@@ -10,17 +10,13 @@ export class ApiRequest {
     });
   }
 
-  async get<T, B>(path: string, query?: { params: B; cancelToken?: CancelToken }): Promise<void | T> {
+  async get<T, B>(path: string, query?: { params: B; cancelToken?: CancelToken }): Promise<T> {
```
atymchuk, line 13: When the promise throws, isn't that the case when it returns `void`?

tiagomotasantos: When it throws the execution moves to the catch block so it doesn't actually return anything. If there's no catch block the browser will handle the error but still the code after won't be executed so there's no need to expect void as return.

Made this snippet to try to explain it: https://stackblitz.com/edit/js-rxurye?file=index.js

fabio-silva (dismissed):

pmm-app/src/shared/components/helpers/api.ts
atymchuk: What I see from the code is that `.catch(e)` statement is always there and it always re-throws the error. This means that, although there is no explicit `return` statement, the engine kicks in and returns `void` (or, more precisely, `unknown` in latest versions of typescript).

I asked because leaving only `Promise<T>` would not correspond to what's happening in reality, which is IMO `Promise<T | unknown>`.

My point is: absence of a return statement does not mean there is no `return`.

My other point is: the `.catch` statement seems to be useless if the only thing it does is re-throwing the error.

tiagomotasantos: The catch is not just re throwing the error, it is showing the error message.

Can you please show an use case where the code after the get will receive `unknown`?

tiagomotasantos: This is basically the definition of `throw`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/throw
_Execution of the current function will stop (the statements after throw won't be executed)_

So the get will either return `T` or it will `throw` it won't ever return `unknown`

atymchuk: > The catch is not just re throwing the error, it is showing the error message.
> 
> Can you please show an use case where the code after the get will receive `unknown`?

This is where. 
https://github.com/percona/grafana-dashboards/pull/1297/files#diff-cd8788d8824d855138634130187cbbb636faa9ca5e17e9fb381f43fc8d38b290R19

pmm-app/src/shared/components/helpers/api.ts
```diff
@@ -42,21 +38,21 @@ export class ApiRequest {
...
 
-  async patch<T, B>(path: string, body: B): Promise<void | T> {
+  async patch<T, B>(path: string, body: B): Promise<T> {
     return this.axiosInstance
       .patch<T>(path, body)
       .then((response): T => response.data)
-      .catch((): void => {
-        // Notify.error(e.message);
+      .catch((e) => {
+        throw e;
```
atymchuk, line 55: Omitting these two lines of code leads to the same result as leaving them here, so the question is: what additional value does it bring?

pmm-app/src/shared/components/helpers/api.ts
tiagomotasantos: That's where it throws, so it does not return. Also that's inside the API helper function, it's not affected by the return type.

pmm-app/src/shared/components/helpers/api.ts
tiagomotasantos: Right! Fixed by show the error notification there to make it consistent with the other functions. Ofc deleting the catch was possible too but solved this way to still get notifications, thanks!!

pmm-app/src/shared/components/helpers/api.ts
tiagomotasantos: If we were putting a wrong return type typescript compiler would warn us.

atymchuk: I think you are right. I was pointing to a wrong solution, something that has to do with type inference which is not what we are trying to do )))

atymchuk (approved):

PR #1315, opened 2022-02-28
PMM-9573 Node 16 update

tiagomotasantos:
Update required dependencies for Node 16

- `node-sass` only works until Node 15. Using `saas` as replacement
- creates custom wrapper for highlight.js and updates it to latest version

https://jira.percona.com/browse/PMM-9573


fabio-silva (approved):

pmm-app/src/shared/components/Hightlight/Highlight.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+export const Highlight = ({ children, className, language }: HighlightProps) => {
+  const styles = useStyles(getStyles);
+  const ref = useRef<HTMLElement>(null);
+
+  useEffect(() => {
+    if (ref.current != null) {
+      highlight.highlightElement(ref.current);
+    }
+  }, [ref]);
```
atymchuk, line 16: It should probably be [ref.current], since `ref` may continue to refer to the same object. Can you double-check?

tiagomotasantos: That was my first thought as well, but "mutable values don't re-render the component" as stated by the following error:

Screenshot 2022-03-03 at 10 25 52

`useEffect` runs after the first render so we are sure that `ref.current` is already populated. In reality we don't even need to check if it's `null`, I'm just doing it to shut up TypeScript since the type of the ref is `HTMLElement` but it's being initialised to `null` and `highlightElement` doesn't accept `null`. Also having ref in the dependencies array or having an empty array is the same for the same reasons.

This would change if we wanted the component to re-render after the ref changes, we would have to use `useState` or `useCallback`, but in this case the content of the ref doesn't change as the HTML element that it points to is always the same.

PR #1394, opened 2022-09-22
PMM-10751 move service name panel to top

anton-bystrov:
https://jira.percona.com/browse/PMM-10751. Move panel to top and hide disclaimer panel.

изображение


dygksy (changes requested):
Disclaimer should be the first row
Other service name row should be placed under Disclaimer row.

anton-bystrov:
> Disclaimer should be the first row Other service name row should be placed under Disclaimer row.

изображение

dygksy (approved):

PR #1417, opened 2022-11-04
Grafana upgrade

fabio-silva:


.github/workflows/build.yml
```diff
@@ -5,20 +5,34 @@ on: pull_request
...
+        uses: percona-platform/checkout@v2
 
-      - name: Build grafana-dashboards
+      - name: Run with Node 16
         uses: percona-platform/setup-node@v2
         with:
-          node-version: "16"
-          cache: "npm"
-          cache-dependency-path: pmm-app/package-lock.json
+          node-version: ${{ matrix.node-version }}
```
atymchuk, line 19: I don't think there is a benefit of using a matrix here, as we probably won't test different versions. Maybe just use the value as before?

pmm-app/src/pmm-qan/panel/QueryAnalytics.tools.ts
```diff
@@ -22,7 +22,7 @@ export function buildShareLink(from: number, to: number) {
 
   params.set('from', `${from}`);
   params.set('to', `${to}`);
-  params.set('orgId', config.bootData.user.orgId);
+  params.set('orgId', `${config.bootData.user.orgId}`);
```
atymchuk, line 25: Is this really different from the one before? Or is it meant to type-convert from Integer to String?

.github/workflows/build.yml
fabio-silva: Sure thing, just changed it 👍

pmm-app/src/pmm-qan/panel/QueryAnalytics.tools.ts
fabio-silva: Yeah, we now get an error while trying to use a number there

matejkubinec (approved):

solovevayaroslavna (approved):

PR #1424, opened 2022-11-21
PMM-10558 Changed query for "Top 5 Collections by Documents Changed" panel

anton-bystrov:
PMM-10558 changed query and added to description text.
image

### DIFF
--- a/dashboards/Experimental/MongoDB_Collection_Details.json
+++ b/dashboards/Experimental/MongoDB_Collection_Details.json
@@ -677,7 +677,7 @@
       "id": 68,
       "links": [],
       "options": {
-        "content": "The next two graphs are available only when `--enable-all-collectors` option is used in pmm-admin.",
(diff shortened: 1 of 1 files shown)


rasika-chivate:
@anton-bystrov I think this is correct: The following two graphs are available only when --enable-all-collectors is used in pmm-admin. The graph **Top 5 Collection by Documents Changed**  displays data only on selecting the Primary node."

denisok (dismissed):

anton-bystrov:
> @anton-bystrov I think this is correct: The following two graphs are available only when --enable-all-collectors is used in pmm-admin. The graph **Top 5 Collection by Documents Changed** displays data only on selecting the Primary node."
> 
> If a non-primary node is used, the data is not displayed, right?

Changed

denisok (approved):

rasika-chivate (approved):

PR #1441, opened 2023-01-20
PMM-11408 Removed multivalue variable for some panels

anton-bystrov:
PMM-11408 Added new hidden variable to prevent "All" option in some links.


ritbl (approved):
LGTM

PR #1454, opened 2023-02-08
PMM-11497 update vertamedia clickhouse data source.

BupycHuk:


atymchuk:
The big question is: now that there is a [clickhouse datasource](https://grafana.com/grafana/plugins/grafana-clickhouse-datasource/) from grafana, do we need to keep using and updating this one?

anton-bystrov (approved):

PR #1457, opened 2023-02-10
PMM-11508 Add welcome block component

0leksii:
Add a block with a welcome introduction and initial
set of action cards for onboarding.

https://user-images.githubusercontent.com/39950859/218132969-f8de3fe0-4fc3-44b3-8e6e-d646d6a4898c.mov


pmm-app/src/onboard/components/ActionCard/ActionCard.tsx
```diff
@@ -0,0 +1,122 @@
...
+        <Card.Description className={styles.cardDescription}>{description}</Card.Description>
+      </Card>
+    </div>
+  );
+};
+
+const getStyles = (theme: GrafanaTheme2) => ({
+  cardHeading: css`
+    font-family: 'Roboto';
+    font-style: normal;
```
dliakhov, line 78: I think we don't need to specify here font-family and font-style here, it's specified already for the html and body tags.

Could you please remove it from othere css rules as well?

pmm-app/src/onboard/components/ActionCard/ActionCard.tsx
```diff
@@ -0,0 +1,122 @@
...
+    line-height: 21px;
+    color: ${theme.colors.text.primary};
+  `,
+  cardDescription: css`
+    min-height: ${theme.typography.body.lineHeight}em
+  `,
+  externalLinkIcon: css`
+    height: 16.67px;
+    width: 16.67px;
+    margin-right: 0 !important;
```
dliakhov, line 90: do we need this? to use !important rule you should have good reason for this, as I know people try to avoid using that.

0leksii: yeah, it is unfortunately set as `!important` in grafana, so to align the icon appropriately had to override it

pmm-app/src/onboard/components/ActionCard/ActionCard.tsx
0leksii: Done

PR #1465, opened 2023-02-27
PMM-11664 Added node_type, environment, region vars.

anton-bystrov:
PMM-11664 Added new variables at all OS panels. Now this story: Region - Node Type - Environment - Node Name

изображение


cezmunsta (approved):
UI seems to be working as expected

PR #1469, opened 2023-03-09
How the dashboard can be converted for use on dedicated Prometheus instances？

z3473:
Hi,How the dashboard can be converted for use on dedicated Prometheus instances?
[root@localhost misc]# ./convert-dash-from-PMM.py ../dashboards/MongoDB/MongoDB_Cluster_Summary.json 
./convert-dash-from-PMM.py: line 19: import: command not found
./convert-dash-from-PMM.py: line 20: import: command not found
./convert-dash-from-PMM.py: line 21: import: command not found
./convert-dash-from-PMM.py: line 24: services: command not found
./convert-dash-from-PMM.py: line 27: syntax error near unexpected token `('
./convert-dash-from-PMM.py: line 27: `def add_datasource_variable(dashboard):'
[root@localhost misc]# python -V
Python 2.7.5
@anton-bystrov


anton-bystrov:
Hello! You should use Python 3 and command like this `python misc/convert-dash-from-PMM.py dashboards/PostgreSQL/PostgreSQL_Instances_Compare.json` i.e. python "path and name of script" "target path and name dashboard"
изображение

PR #1475, opened 2023-03-16
PMM-10698 Fixed panels at Node Summary row

anton-bystrov:
PMM-10698


rnovikovP:
Let's do a basic sanity review here.
It's clear - the formulas need cleaning up  (expected to be done in Container and k8s monitoring task)
For now the main goal - fix the problem with different  data for node metrics in Dashboard and details row

BupycHuk (commented):
Hard to review, but probably we should try to rely on node type != "container" for non-container instances to ignore node_exporter metrics from containers

BupycHuk (approved):

PR #1487, opened 2023-04-07
PMM-10799 Show aliased table

IuriiMr:
PMM-10799

Feature Build:
 - https://github.com/Percona-Lab/pmm-submodules/pull/3206


pmm-app/src/pmm-qan/panel/components/Details/Table/TableContainer.hooks.ts
```diff
@@ -11,14 +11,16 @@ export const useTables = (example, explains, databaseType): any[] => {
     const getTables = async () => {
       setLoading(true);
       setTables([]);
-      if (databaseType === Databases.mysql && jsonExplain.value) {
-        const parsedJSON = JSON.parse(jsonExplain.value);
+      if (databaseType === Databases.mysql && example && jsonExplain) {
+        // eslint-disable-next-line global-require
+        const parser = require('sqlite-parser');
```
dliakhov, line 16: could be this imported like other libs (like `import { get } from 'lodash';`) to avoid importing this library each time when this code is called?

pmm-app/src/pmm-qan/panel/components/Details/Table/TableContainer.hooks.ts
```diff
@@ -11,14 +11,16 @@ export const useTables = (example, explains, databaseType): any[] => {
...
+        const ast = parser(example?.example);
 
-        const realTableName = [
-          get(parsedJSON, 'real_table_name'),
+        const tablesResult = [
+          get(ast, 'statement.0.from.source.name') || get(ast, 'statement.0.from.name'),
```
dliakhov, line 20: do we need to check whether `statement` has non empty array? could be here the case if `statement` is empty or null?

IuriiMr: in general, if we get an example and jsonExplain we will have a statement with the proper query.

pmm-app/src/pmm-qan/panel/components/Details/Table/TableContainer.hooks.ts
ritbl: I have another question, is it a workaround of somesort?

pmm-app/package.json
```diff
@@ -66,6 +66,7 @@
     "scroll-into-view": "^1.14.2",
     "simplebar-react": "^3.0.0-beta.8",
     "sql-formatter": "^2.3.3",
+    "sqlite-parser": "^1.0.1",
```
ritbl, line 69: why do we need that?

pmm-app/src/pmm-qan/panel/components/Details/Table/TableContainer.hooks.ts
```diff
@@ -11,14 +11,16 @@ export const useTables = (example, explains, databaseType): any[] => {
     const getTables = async () => {
       setLoading(true);
       setTables([]);
-      if (databaseType === Databases.mysql && jsonExplain.value) {
-        const parsedJSON = JSON.parse(jsonExplain.value);
+      if (databaseType === Databases.mysql && example && jsonExplain) {
+        // eslint-disable-next-line global-require
+        const parser = require('sqlite-parser');
+        const ast = parser(example?.example);
```
ritbl, line 17: So `example` can be empty? What is an expected behavour

ritbl (changes requested):
please the see comments

PR #1491, opened 2023-04-16
PMM-9844 Fix details classic explain table view

IuriiMr:
PMM-9844

Feature Build:
 - https://github.com/Percona-Lab/pmm-submodules/pull/3218

### DIFF
--- a/pmm-app/src/pmm-qan/panel/components/Details/Explain/Explain.tools.ts
+++ b/pmm-app/src/pmm-qan/panel/components/Details/Explain/Explain.tools.ts
@@ -28,7 +28,7 @@ export const processClassicExplain = (classic): ClassicExplainInterface => {
     .map((e) => (String(e) ? e.trim() : ''))
     .filter(Boolean)
     .reduce((acc, row, index) => {
-      acc[headerList[index].accessor] = row.split(',').join(', ');
+      acc[headerList[index].accessor] = row.replaceAll(',', ', ');
 
       return acc;
     }, {}));
--- a/pmm-app/src/pmm-qan/panel/components/Details/Explain/Explain.tools.ts
+++ b/pmm-app/src/pmm-qan/panel/components/Details/Explain/Explain.tools.ts
@@ -28,7 +28,7 @@ export const processClassicExplain = (classic): ClassicExplainInterface => {
     .map((e) => (String(e) ? e.trim() : ''))
     .filter(Boolean)
     .reduce((acc, row, index) => {
-      acc[headerList[index].accessor] = row.replaceAll(',', ', ');
+      acc[headerList[index].accessor] = row;
 
       return acc;
     }, {}));
--- a/pmm-app/src/pmm-qan/panel/components/Details/Explain/Explain.tools.ts
+++ b/pmm-app/src/pmm-qan/panel/components/Details/Explain/Explain.tools.ts
@@ -28,7 +28,7 @@ export const processClassicExplain = (classic): ClassicExplainInterface => {
     .map((e) => (String(e) ? e.trim() : ''))
     .filter(Boolean)
     .reduce((acc, row, index) => {
-      acc[headerList[index].accessor] = row;
+      acc[headerList[index].accessor] = row.split(',').join(', ');
 
       return acc;
     }, {}));
--- a/pmm-app/src/pmm-qan/panel/components/Details/Explain/components/ClassicExplain/ClassicExplain.tsx
+++ b/pmm-app/src/pmm-qan/panel/components/Details/Explain/components/ClassicExplain/ClassicExplain.tsx
@@ -2,13 +2,17 @@ import React from 'react';
 import { Scrollbar } from 'shared/components/Elements/Scrollbar/Scrollbar';
 import { Overlay } from 'shared/components/Elements/Overlay/Overlay';
 import { Table } from 'shared/components/Elements/Table';
+import { useTheme } from '@grafana/ui';
 import { processClassicExplain } from '../../Explain.tools';
 import { Messages } from '../../../Details.messages';
 import { ReplacedQueryMessage } from '../ReplacedQueryMessage/ReplacedQueryMessage';
+import { getStyles } from '../../../../../../../shared/components/Elements/Table/Table.styles';
 
 export const ClassicExplain = ({ classicExplain }) => {
   const { value: explain } = classicExplain;
   const processedExplain = processClassicExplain(explain?.explain_result);
+  const theme = useTheme();
+  const styles = getStyles(theme);
 
   return (
     <Overlay isPending={classicExplain.loading}>
@@ -19,6 +23,7 @@ export const ClassicExplain = ({ classicExplain }) => {
         && processedExplain.rows.length ? (
           <div data-testid="classic-explain-value">
             <Table
+              className={styles.tableClassicExplain}
               columns={processedExplain.columns}
               data={processedExplain.rows}
               noData={null}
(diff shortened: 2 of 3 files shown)


pmm-app/src/pmm-qan/panel/components/Details/Explain/Explain.tools.ts
```diff
@@ -28,7 +28,7 @@ export const processClassicExplain = (classic): ClassicExplainInterface => {
     .map((e) => (String(e) ? e.trim() : ''))
     .filter(Boolean)
     .reduce((acc, row, index) => {
-      acc[headerList[index].accessor] = row;
+      acc[headerList[index].accessor] = row.split(',').join(', ');
```
pkadej, line 31: How about replaceAll method?
```
acc[headerList[index].accessor] = row.replaceAll(',', ', ');
```

IuriiMr: The same result but it seems to be faster.
Thanks.

pkadej (dismissed):

pmm-app/src/pmm-qan/panel/components/Details/Explain/Explain.tools.ts
```diff
@@ -28,7 +28,7 @@ export const processClassicExplain = (classic): ClassicExplainInterface => {
     .map((e) => (String(e) ? e.trim() : ''))
     .filter(Boolean)
     .reduce((acc, row, index) => {
-      acc[headerList[index].accessor] = row;
+      acc[headerList[index].accessor] = row.replaceAll(',', ', ');
```
ritbl, line 31: this kind of workaround should be explained. I don't think it is a proper fix. Why do you just use css?

ritbl (changes requested):
see my comments

pmm-app/src/pmm-qan/panel/components/Details/Explain/Explain.tools.ts
IuriiMr: The initial/default value of a CSS property 'white-space' is 'normal', so we do not need to change it.
With this value inside the elements the lines are broken by space as necessary to fill the element (the cell in this case).
Resolving the second part of the ticket,- adding spaces after commas,- works for the text overflow as well.

IuriiMr:
Fixed by adding CSS wrap for the classic explain table.
Possible sequences of white space (in case of wrong formatted response) are collapsed with CSS 'white-space: normal'.

pmm-app/src/pmm-qan/panel/components/Details/Explain/components/ClassicExplain/ClassicExplain.tsx
```diff
@@ -2,13 +2,17 @@ import React from 'react';
 import { Scrollbar } from 'shared/components/Elements/Scrollbar/Scrollbar';
 import { Overlay } from 'shared/components/Elements/Overlay/Overlay';
 import { Table } from 'shared/components/Elements/Table';
+import { useTheme } from '@grafana/ui';
 import { processClassicExplain } from '../../Explain.tools';
 import { Messages } from '../../../Details.messages';
 import { ReplacedQueryMessage } from '../ReplacedQueryMessage/ReplacedQueryMessage';
+import { getStyles } from '../../../../../../../shared/components/Elements/Table/Table.styles';
```
ritbl, line 9: what about absolute import?

pmm-app/src/shared/components/Elements/Table/Table.styles.ts
```diff
@@ -66,5 +66,14 @@ export const getStyles = stylesFactory((theme: GrafanaTheme) => {
     checkboxColumn: css`
       width: 20px;
     `,
+    tableClassicExplain: css`
+    table {
+      td {
+        white-space: normal;
+        max-inline-size: 550px;
```
ritbl, line 73: I don't think it is good idea to have the same max size. Each column should have it's own max size, and in %.

pmm-app/src/pmm-qan/panel/components/Details/Explain/Explain.tools.ts
ritbl: I propose not to modify that.

pmm-app/src/shared/components/Elements/Table/Table.styles.ts
```diff
@@ -66,5 +66,14 @@ export const getStyles = stylesFactory((theme: GrafanaTheme) => {
     checkboxColumn: css`
       width: 20px;
     `,
+    tableClassicExplain: css`
+    table {
+      td {
+        white-space: normal;
```
ritbl, line 72: why do we need this? why not `word-break: break-all;`

pmm-app/src/pmm-qan/panel/components/Details/Explain/components/ClassicExplain/ClassicExplain.tsx
IuriiMr: Fixed.

pmm-app/src/shared/components/Elements/Table/Table.styles.ts
IuriiMr: We do not need it really, just tried to make clear that this property could be used.
word-break: break-word could be useful to keep words as far but break-all works well in any case.
Fixed.

pmm-app/src/shared/components/Elements/Table/Table.styles.ts
IuriiMr: I propose to use 'min-inline-size' instead to avoid linebraking with the 'NULL' and to use maximum possible size for the widthest columns.

pmm-app/src/pmm-qan/panel/components/Details/Explain/Explain.tools.ts
IuriiMr: Fixed.

pmm-app/src/shared/components/Elements/Table/Table.styles.ts
```diff
@@ -66,5 +66,13 @@ export const getStyles = stylesFactory((theme: GrafanaTheme) => {
     checkboxColumn: css`
       width: 20px;
     `,
+    tableClassicExplain: css`
+    table {
+      td {
+        min-inline-size: 4rem;
```
ritbl, line 72: why not min width?

pmm-app/src/shared/components/Elements/Table/Table.styles.ts
```diff
@@ -66,5 +66,13 @@ export const getStyles = stylesFactory((theme: GrafanaTheme) => {
     checkboxColumn: css`
       width: 20px;
     `,
+    tableClassicExplain: css`
```
ritbl, line 69: name `tableClassicExplain` is not clear, what does it mean classic explain?

ritbl (changes requested):
please see my comments

pmm-app/src/shared/components/Elements/Table/Table.styles.ts
IuriiMr: 'Explain' as part of the query selected, as listed in the summary table, consists of 'Classic' and 'JSON' parts. As I've been working at the table from 'Classic' part of 'Explain' tab, I chose the naming to follow this belongings: 'tableClassicExplain'.

pmm-app/src/shared/components/Elements/Table/Table.styles.ts
IuriiMr: The min-inline-size property defines the horizontal or vertical minimal size of an element's block, depending on its writing mode. It corresponds to either the min-width or the min-height property, depending on the writing-mode. 
So, I chose it to set min 4rem for the direction of the lines of text to fit the string 'NULL' as the word that could be shown without line braking.  As the 'NULL' could be appeared in any column we do not need to set a minimum for each column separately.

pmm-app/src/shared/components/Elements/Table/Table.styles.ts
IuriiMr: Fixed with name changing and adding comments.

pmm-app/src/shared/components/Elements/Table/Table.styles.ts
```diff
@@ -66,5 +66,15 @@ export const getStyles = stylesFactory((theme: GrafanaTheme) => {
     checkboxColumn: css`
       width: 20px;
     `,
+    tableCellWrap: css`
```
ritbl, line 69: I'd go with `/* Wraps long cell content  */`

PR #1494, opened 2023-04-26
PMM-11968 Add query comments to the filters

IuriiMr:
PMM-11968

Feature Build:
 - https://github.com/Percona-Lab/pmm-submodules/pull/3235


pmm-app/src/pmm-qan/panel/components/Filters/hooks/useFilters.ts
IuriiMr, line 1: Filtering the response from BE by labels from FILTERS_GROUPS and HIDDEN_FILTER_LABELS to keep only label from key='value' comments from the queries. After that adding additional group from comment key to the FILTERS_GROUP to follow the way for the filtering by group labels.

pmm-app/src/pmm-qan/panel/components/Filters/Filters.constants.ts
IuriiMr, line 1: COMMENT_NAME_LENGTH is the value of the string length that could be presenting in one line in filters section.
HIDDEN_FILTER_LABELS are the labels from BE response that do not need to add to the filters.

pmm-app/src/pmm-qan/panel/components/Filters/Filters.tsx
```diff
@@ -26,13 +26,13 @@ export const Filters: FC = () => {
     contextActions,
     panelState: { loadingDetails, rawTime },
   } = React.useContext(QueryAnalyticsProvider);
-  const [filters, loading] = useFilters();
+  const [filters, loading, FILTERS_GROUPS] = useFilters();
```
0leksii, line 29: Just wondering - what is the purpose of passing a const `FILTERS_GROUPS` via the hook, rather than the way it was done previously just importing it? As I understood it is still const and values do not change in there?
Also, considering that it is coming from the hook, I would not suggest using all-caps naming here as it is quite misleading and makes you think that `FILTERS_GROUPS` is some sort of the constant but it is not.

pmm-app/src/pmm-qan/panel/components/Filters/hooks/useFilters.ts
```diff
@@ -19,6 +24,23 @@ export const useFilters = (): [any, boolean, boolean] => {
...
+            return obj;
+          }, {});
+
+        Object.keys(filteredResult).forEach((commentKey) => FILTERS_GROUPS.push({
+          name: commentKey
+            .replace(/^\w/, (c) => c.toUpperCase())
+            .replace(/_/g, ' ')
+            .substring(0, COMMENT_NAME_LENGTH),
+          dataKey: commentKey,
+        }));
```
0leksii, line 42: Again just a suggestion, but I would merge those two:
```suggestion
        const filtersGroups = Object.keys(result)
          .filter((key) => !FILTERS_GROUPS.some((group) => group.dataKey === key))
          .filter((key) => !HIDDEN_FILTER_LABELS.includes(key))
          .reduce((groups, key) => {
            groups.push({
              name: key
               .replace(/^\w/, (c) => c.toUpperCase())
               .replace(/_/g, ' ')
               .substring(0, COMMENT_NAME_LENGTH),
             dataKey: key,
            });

            return groups;
          }, [... FILTERS_GROUPS]);
```

by that you will omit pushing anything directly to the `const FILTERS_GROUPS` as well as less iteration on the result keys.
and then you can just return `filtersGroups`

pmm-app/src/pmm-qan/panel/components/Filters/Filters.tools.ts
```diff
@@ -1,11 +1,14 @@
 import { getTemplateSrv } from '@grafana/runtime';
-import { FILTERS_GROUPS } from './Filters.constants';
 
-export const getSelectedCheckboxes = (filters) => FILTERS_GROUPS.map((group) => filters[group.dataKey])
-  .filter(Boolean)
-  .map((item) => item.name)
-  .reduce((acc, item) => acc.concat(item), [])
-  .some((item) => item.checked);
+export const getSelectedCheckboxes = (filters, FILTERS_GROUPS) => {
+  const selectedFilters = FILTERS_GROUPS.map((group) => filters[group.dataKey])
```
0leksii, line 4: again here - I would not use the misleading naming that indicate that value is a const

IuriiMr: Fixed with changing the naming.

pmm-app/src/pmm-qan/panel/components/Filters/hooks/useFilters.ts
IuriiMr: Sorry for misleading naming, but the idea was to add the comments (as result of the filtering the response by FILTERS_GROUP and HIDDEN_FILTERS_LABELS) to the existing FILTERS_GROUP.
Fixed with changing the naming.

0leksii (dismissed):

0leksii (approved):

pmm-app/src/pmm-qan/panel/components/Filters/Filters.constants.ts
```diff
@@ -4,9 +4,22 @@ import { getServiceType } from './Filters.tools';
 export const FILTERS_BODY_HEIGHT = 600;
 export const FILTERS_HEADER_SIZE = 50;
 export const FILTERS_MARGIN_BOTTOM = 20;
+export const COMMENT_NAME_LENGTH = 24;
```
ritbl, line 7: it is really not clear, how about using "<comment-str>".length?

pmm-app/src/pmm-qan/panel/components/Filters/Filters.tools.ts
```diff
@@ -1,11 +1,14 @@
 import { getTemplateSrv } from '@grafana/runtime';
-import { FILTERS_GROUPS } from './Filters.constants';
 
-export const getSelectedCheckboxes = (filters) => FILTERS_GROUPS.map((group) => filters[group.dataKey])
-  .filter(Boolean)
-  .map((item) => item.name)
-  .reduce((acc, item) => acc.concat(item), [])
-  .some((item) => item.checked);
+export const getSelectedCheckboxes = (filters, filtersGroups) => {
+  const selectedFilters = filtersGroups.map((group) => filters[group.dataKey])
```
ritbl, line 4: what if `group` is null/undefined or there is no item in filters for `group.dataKey`

pmm-app/src/pmm-qan/panel/components/Filters/hooks/useFilters.ts
```diff
@@ -19,7 +25,23 @@ export const useFilters = (): [any, boolean, boolean] => {
       try {
         const result = await FiltersService.getQueryOverviewFiltersList(labels, from, to, columns[0]);
 
+        const filtersGroups = Object.keys(result)
+          .filter((key) => !FILTERS_GROUPS.some((group) => group.dataKey === key))
```
ritbl, line 29: it would be good to clarify what you are filtering out

pmm-app/src/pmm-qan/panel/components/Filters/hooks/useFilters.ts
```diff
@@ -19,7 +25,23 @@ export const useFilters = (): [any, boolean, boolean] => {
...
+          .filter((key) => !HIDDEN_FILTER_LABELS.includes(key))
+          .reduce((groups, key) => {
+            groups.push({
+              name: key
```
ritbl, line 33: it would be good to refactor this code so that it is more readable

ritbl (dismissed):

pmm-app/src/pmm-qan/panel/components/Filters/Filters.tools.ts
IuriiMr: This `filtersGroups` is based on the const FILTERS_GROUPS with the addition of groups from the comment of the selected query (if there is one), so it always has groups from the constant.

0leksii (approved):

pmm-app/src/pmm-qan/panel/components/Filters/hooks/useFilters.ts
```diff
@@ -19,7 +25,25 @@ export const useFilters = (): [any, boolean, boolean] => {
       try {
         const result = await FiltersService.getQueryOverviewFiltersList(labels, from, to, columns[0]);
 
+        const filtersGroups = Object.keys(result)
+          .filter((key) => !FILTERS_GROUPS.some((group) => group.dataKey === key))
+          .filter((key) => !HIDDEN_FILTER_LABELS.includes(key))
+          .reduce((groups, key) => {
+            if (result[key].name) {
+              groups.push({
+                name: key
```
matejkubinec, line 34: Could you extract this to an util function with a descriptive name?

Otherwise looks good.

fabio-silva (approved):

PR #1544, opened 2023-11-20
PMM-12425 fix infinite loop for setting timerange

YashSartanpara1:
[PMM-12425 ](https://jira.percona.com/browse/PMM-12425) - QAN picking different date range will cause an error

qan2

Screenshot 2023-11-22 123534

Previous fixes should be tested after this fix -
[PMM-9987](https://jira.percona.com/browse/PMM-9987)
[PMM-10483](https://jira.percona.com/browse/PMM-10483)

### DIFF
--- a/pmm-app/src/pmm-qan/panel/provider/provider.tsx
+++ b/pmm-app/src/pmm-qan/panel/provider/provider.tsx
@@ -210,17 +210,8 @@ export const UrlParametersProvider = (props) => {
   const [previousState, setPreviousState] = useState(panelState);
 
   useEffect(() => {
-    const newTo = getAbsoluteTime(timeRange.raw.to);
-
-    if (newTo === 'now') {
-      setToTimeMomentValue(timeRange.to.subtract(1, 'minute')
-        .format('YYYY-MM-DDTHH:mm:ssZ'));
-      setFromTimeMomentValue(timeRange.from.subtract(1, 'minute')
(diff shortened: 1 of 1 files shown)


pmm-app/src/pmm-qan/panel/provider/provider.tsx
```diff
@@ -215,8 +216,13 @@ export const UrlParametersProvider = (props) => {
     if (newTo === 'now') {
       setToTimeMomentValue(timeRange.to.subtract(1, 'minute')
         .format('YYYY-MM-DDTHH:mm:ssZ'));
-      setFromTimeMomentValue(timeRange.from.subtract(1, 'minute')
-        .format('YYYY-MM-DDTHH:mm:ssZ'));
+
+      if (moment.isMoment(timeRange.raw.from)) {
```
matejkubinec, line 220: What was exactly the issue?

YashSartanpara1: https://jira.percona.com/browse/PMM-9987 
as per above ticket minus 1 minute was only required when the 'From' and 'To' time is relative (ie ranges like now-6h, now-30m etc.), Hence the special case is required when the 'From' time is absolute and To time is relative 'now'. Previous logic was doing minus 1 minute from the absolute 'From' time which was causing infinite loop.

matejkubinec (approved):

BupycHuk (approved):

PR #1546, opened 2023-12-06
PMM-12284 remove Alertmanager datasource

ademidoff:


matejkubinec (commented):
@ademidoff 

Could you check `misc/convert-dash-to-PMM.py` there seems to be a mention of the datasource.

PR #1559, opened 2024-01-24
PMM-12866 update k8s db cluster

YashSartanpara1:


rnovikovP (approved):

ademidoff:
@YashSartanpara1 I don't know if it's a good idea to use spaces in a directory name. Does this PR build OK, i.e. no errors? Is there a feature build to prove that?

rnovikovP:
> @YashSartanpara1 I don't know if it's a good idea to use spaces in a directory name. Does this PR build OK, i.e. no errors? Is there a feature build to prove that?

this is the current folder so we already have space here. i think we can keep it for now . when we;ll GA - the space will go away

ademidoff (approved):

PR #1566, opened 2024-02-15
PMM-12888 Implemented new Patroni dashboard.

guriandoro:
New dashboard as showcased in https://www.percona.com/blog/monitoring-a-postgresql-patroni-cluster/.

There are some minor differences with the one shown in the blog, like separating Patroni and Postrges configurations/panels. I think it's cleaner in this way. The WAL location graphs have hidden rate queries, which can be helpful in some cases, but also add a bit of noise. I was not sure what to do with them, so I decided to leave them disabled by default.

Let me know what you think about it. 

Thanks!

FB: https://github.com/Percona-Lab/pmm-submodules/pull/3546 

https://perconadev.atlassian.net/browse/PMM-12888


dashboards/Experimental/Patroni_Details.json
```diff
@@ -0,0 +1,3766 @@
...
+        "datasource": "Metrics",
+        "definition": "label_values({__name__=~\"pg_up|mysql_up|mongodb_up|proxysql_mysql_status_active_transactions\"}, replication_set)",
+        "hide": 2,
+        "includeAll": true,
+        "label": "Replication Set",
+        "multi": true,
+        "name": "replication_set",
+        "options": [],
+        "query": {
+          "query": "label_values({__name__=~\"pg_up|mysql_up|mongodb_up|proxysql_mysql_status_active_transactions\"}, replication_set)",
```
ademidoff, line 3579: I wonder if querying for anything other than `pg_up` is of relevance here?

dashboards/Experimental/Patroni_Details.json
```diff
@@ -0,0 +1,3766 @@
...
+      "12h",
+      "24h",
+      "2d",
+      "7d",
+      "30d"
+    ]
+  },
+  "timezone": "",
+  "title": "Patroni Details",
+  "uid": "rLzu8z_Vk",
```
ademidoff, line 3763:
```suggestion
  "uid": null,
```
FYI: we should not commit the IDs - for provisioned dashboards those get generated by grafana at the moment of provisioning.

dashboards/Experimental/Patroni_Details.json
guriandoro: Great question. I was not using this, and just opted for:

Environment / Cluster / Service Name / Scope Name

However, the `pmm-admin add external` command offers the `--replication-set` argument, so we could add it. The postgresql exporter also has this argument available, so this will boil down to how we want to do it for Postgres in general.

Let me know and I can do the changes.

rnovikovP: Let's keep it as is for Experimental. 
If this is what I'm thinking - it's the "how to get the list of Replicas in general," and for now, we use the metrics, but when our inventory becomes more reliable ("labels auto discovery") - we must move it to be based on inventory, probably.

rnovikovP (approved):

PR #1571, opened 2024-02-22
PMM-12961 [Tech Preview] New Databases Overview Dashboard

nethalo:
[PMM-12961](https://perconadev.atlassian.net/browse/PMM-12961)
New experimental GLOBAL overview dash

[PMM-12961]: https://perconadev.atlassian.net/browse/PMM-12961?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

image


rnovikovP (approved):

dashboards/Experimental/databases_overview.json
```diff
@@ -0,0 +1,1055 @@
...
+          "type": "dashboard"
+        },
+        "type": "dashboard"
+      }
+    ]
+  },
+  "editable": true,
+  "fiscalYearStartMonth": 0,
+  "graphTooltip": 1,
+  "id": 78,
```
BupycHuk, line 27: can we drop `id` fields here? Here and in subpanels

BupycHuk (approved):

BupycHuk:
@rnovikovP what do you think should it be called `QAN overview` or something else, as far as I see it uses only QAN data and don't use exporters metrics.

dashboards/Experimental/databases_overview.json
ademidoff, line 27:
```suggestion
  "id": null,
```

ademidoff: @nethalo You can just replace the integers with `null`, so in this case it'd become `"id": null,`

rnovikovP:
> @rnovikovP what do you think should it be called `QAN overview` or something else, as far as I see it uses only QAN data and don't use exporters metrics.

It's not only QAN. CH datasource here because Metrics is default and removed.  
The final name - will be set when we do a GA after feedback.

PR #1572, opened 2024-02-27
PMM-9403 Fix replication lag panel.

BupycHuk:
Other links:
https://github.com/Percona-Lab/pmm-submodules/pull/3566
https://github.com/percona/mongodb_exporter/pull/804


dashboards/MongoDB/MongoDB_ReplSet_Summary.json
```diff
@@ -378,7 +378,7 @@
         {
           "datasource": "Metrics",
           "exemplar": false,
-          "expr": "avg by (set) (max_over_time(mongodb_mongod_replset_member_replication_lag{set=\"$replset\",state=\"SECONDARY\",service_name=~\"$service_name\"}[${__range}]))",
+          "expr": "avg by (set) (max_over_time(mongodb_mongod_replset_member_replication_lag{set=\"$replset\",state=\"SECONDARY\",self=~\"|1\",service_name=~\"$service_name\"}[${__range}]))",
```
ademidoff, line 381: I have a question: if `self` can take values of either `""` or `"1"`, then would this PR change anything? otherwise, what other values could it take?

BupycHuk: it's made to still show something for old clients which has no value for this label, for new clients it will be "0" or "1" and we should ignore the ones that has "0" as value

ademidoff (approved):

PR #1626, opened 2024-08-30
PMM-13327 - New MongoDB Router Summary dashboard

sleto-it:
- https://perconadev.atlassian.net/browse/PMM-13327

**Still todo:**
- [x] Change plugin from `natel-discrete-panel` to `state-timeline`. 
- [x]  Fix `title` of the first chart (and if needed check `repeat for`). It should be "Overview - name here" but the name is currently not shown.
   - We are currently using `"title": "Overview - $service_name"`. But `$service_name` is not printed. If this is changed with `"title": "Overview - $node_name"` it works in my environment (does it cover all environments? if yes, good.)
  - Should we instead keep the title as it is, and use "repeat for service name"? right now it is using "repeat for node name". Note that if we repeat by service, the dashboard prints too many things. Either one or the other needs fix, I believe. 
  - Fixed name with https://github.com/percona/grafana-dashboards/pull/1626/commits/3b2f2cdca5da47bb62a463f688f2e550ff0feba0
- [x]  Shall we remove "Disk space utilization" (`mongodb_dbstats_fsUsedSize`) ? No data is being reported in my test. I think this metric is only for conf / shard (non arbiters), and does not apply to mongoS
- [x]  Shall we consider Blue color for disk IOPS and network traffic to be similar to replicaset summary dashboard?

### DIFF
--- a/dashboards/MongoDB/MongoDB_Router_Summary.json
+++ b/dashboards/MongoDB/MongoDB_Router_Summary.json
@@ -1807,8 +1807,8 @@
     ]
   },
   "timezone": "",
-  "title": "MongoDB Router Summary (New)",
-  "uid": "mongodb-replset-summary-new2",
+  "title": "MongoDB Router Summary",
+  "uid": "mongodb-router-summary",
   "version": 16,
   "weekStart": ""
 }
--- a/dashboards/MongoDB/MongoDB_Router_Summary.json
+++ b/dashboards/MongoDB/MongoDB_Router_Summary.json
@@ -189,7 +189,7 @@
       "panels": [],
       "repeat": "node_name",
       "repeatDirection": "h",
-      "title": "Overview - $service_name",
+      "title": "Overview - $node_name",
       "type": "row"
     },
     {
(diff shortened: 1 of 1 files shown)


sleto-it:
@YashSartanpara1 FYI

sleto-it:
Hi @igroene when you have a moment, wanted to ask your opinion. See my question above:

```
Shall we remove "Disk space utilization" (mongodb_dbstats_fsUsedSize) ? No data is being reported in my test. I think this metric is only for conf / shard (non arbiters), and does not apply to mongoS
```

WDYT on this topic. Thanks

igroene:
in my test env, mongodb_dbstats_fsUsedSize is populated even on mongos routers

sleto-it:
@igroene thanks.  Do you have this mongoS colocated with other nodes, perhaps this is the reason why you see data? or are you using the terraform script? from a teoretical point of view, which data should we measure for mongoS?   for arbiters (in the replica set summary dashboard) I see the same "no data".

Which MongoDB version are you using? I am on 6.0.16

Thanks!

igroene:
Hey Santo, this is using the terraform stuff so mongos router is on dedicated machines. For arbiters it makes sense to have no data, as dbstats cannot be run on them but mongos should show you data. Probably from mongos perspective the disk space is not that important so let's remove for now it if it's blocking progress. We can always go to the Node Summary on OS folder to check disk space

sleto-it:
thanks for this clarification.  I vote it to keep it as it works for you.

YashSartanpara1 (approved):

PR #1649, opened 2025-01-21
OS/Network_Details: Fix TCP counters

btriller:
* Show rate for counter metrics
* Show established gauge on own axis


YashSartanpara1:
Hi, We are using v3 branch for active development, can we please change the base to percona:v3

PR #1655, opened 2025-02-12
PMM-7 Update compose file for PMM 3

matejkubinec:


pmm-app/docker-compose.yaml
```diff
@@ -24,19 +30,19 @@ services:
 
   pmm-client:
     container_name: pmm-client
-    image: centos
+    image: rockylinux:9
```
ademidoff, line 33: I suggest we use oraclelinux:9 instead, which is the OS we support and use across the board, including the CI environment.

matejkubinec: Good point, updated.

ademidoff (approved):

doracretu3pillar (approved):

PR #1676, opened 2025-04-17
PMM-13940 Change in metricsQL queries for reads/writes

guriandoro:
It was not using filtering labels, so the page size was not being picked up in the multiplication between the two vectors.


guriandoro:
Hi @YashSartanpara1,

Were you able to check this?

PR #1687, opened 2025-06-07
PMM-14072 Automate Grafana plugin upgrade

ademidoff:


misc/Dockerfile.upgrade
```diff
@@ -0,0 +1,8 @@
+FROM grafana/grafana:latest
```
matejkubinec, line 1: Might be good to have an option to specify the grafana version (we don't update all the time to latest, and there may be possibility of some version mismatch - altought small)

ademidoff: Sure, done!

matejkubinec (approved):

misc/Dockerfile.upgrade
ademidoff: Just to be clear: this image is only used as a means of running the CLI to install the plugins, i.e. `grafana cli plugins install ...`.

PR #1688, opened 2025-06-08
PMM-7 Migrate to codecov action

ademidoff:
**Motivation**

The codecov upload has been unstable for quite a long time. The reason is probably that the `codecov` package has been deprecated and archived since 2003.

**What**

This PR migrates to using the Codecov Github workflow, which handles the upload of code coverage artifacts as before. It also delivers a few logical improvements for the Makefile targets and provides a general code cleanup.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,9 +1,10 @@
-name: Test pipeline
+name: Test
 
 on: pull_request
 
 jobs:
   build:
+    name: Build
     runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


matejkubinec (approved):
Looks good, there's a ticket for fixing the code coverage - https://perconadev.atlassian.net/browse/PMM-13705

Also, it seems due to the change in we will need to update required checks (after merge).

YashSartanpara1 (approved):

PR #1693, opened 2025-06-16
PMM-14059 fix service_name in MongoDB cluster summary expression

BupycHuk:


YashSartanpara1:
The query is correct, 

the variable got removed in one of the previous change
query for service_name_mongos variable - 
label_values(mongodb_mongos_sharding_shards_total{environment=~\"$environment\",cluster=~\"$cluster\"}, service_name)

BupycHuk:
@YashSartanpara1 do you mean that new query is correct or the old one?

YashSartanpara1:
I have added the fix, old query was correct but the variable was missing.

matejkubinec (approved):

PR #1711, opened 2025-08-20
PMM-7 Refresh connection section

BupycHuk:
The 'refresh' value controls when the variable is updated: 1=on dashboard load, 2=on time range change. Here, 2 ensures the uptime, connection counts, and active connections variables are refreshed when the time range changes, which is important for accurate reporting. See Grafana docs: https://grafana.com/docs/grafana/latest/variables/variable-types/add-query-variable/#refresh-options

Fixes https://github.com/percona/pmm/issues/4357


ademidoff:
@YashSartanpara1 PTAL

PR #1724, opened 2025-10-07
PMM-14363 - OS Disk Space Details is difficult to interpret

igroene:
Cleanup unused variables
Changed to proper units
Fix group expressions for device and fstype


dashboards/OS/Disk_Details.json
```diff
@@ -55,6 +55,7 @@
   "editable": false,
   "fiscalYearStartMonth": 0,
   "graphTooltip": 1,
+  "id": 21,
```
ademidoff, line 58:
```suggestion
  "id": null,
```

PR #1787, opened 2026-03-30
PMM-14360 Disk I/O Total output is incorrect

igroene:
Fixes the expression in every dashboard


ademidoff (changes requested):
Please do not merge - we will move it to the monorepo this week :)

